PR #21, opened 2015-09-21
Added check to ensure disk exists in both volumes and mounts

bridadan:
When running mbedls quickly (with mbeds unmounting and remounting in the background), sometimes a disk listed as a volume is not yet mounted.

This PR removes any mbed that is not yet mounted from the returned list.


bridadan:
@PrzemekWirkus I see some test have failed, but it looks like this is an issue with the tests, not my changes. Any chance this can be merged soon?

PrzemekWirkus:
@bridadan If there is an issue with tests, please fix them (tests)! :)

PrzemekWirkus:
@bridadan Please make sure your PR will pass tests with:
```python setup.py test```

Issue #25, opened 2015-10-25
Why is the Linux support Ubuntu specific?


jrobeson:
The code seems pretty generic for linux, but maybe I missed something.

simonqhughes:
ARM Internal Ref: IOTSYST-476

PrzemekWirkus:
Hi Johnny,
For Ubuntu Linux distro we have a proper port we've developed, we maintain and, test on our infrastructure.

Nothing stops us as a community from adding a new Linux distro support to mbed-ls's code base but we have to remember that we then have to commit to support and test. It is all about working software others can use seamlessly in the end.

We are still improving our infrastructure and this is why we are currently concentrating (and limiting) our efforts of limited number of supported platforms.

If anyone wants to port mbed-ls they can, and to verify their port they can use packages's unit tests:
```
$ git clone <mbed-ls>
$ cd mbed-ls
... Make your changes ...
$ python setup.py test
```

jrobeson:
isn't it recommended to use udisks api on ubuntu instead of calling the programs directly?

jrobeson:
that's another issue though. I will file that separately.

PrzemekWirkus:
Cool, you are also more than welcome to send us pull request using udisks on Ubuntu to replace current implementation!

PrzemekWirkus:
We have promising PR https://github.com/ARMmbed/mbed-ls/pull/26 to fix this issue.

PrzemekWirkus:
LInux support was merged to master. We will have release soon! :)

PR #26, opened 2015-11-03
Add Linux generic support

dubek:
The new Linux generic support is actually the original code from `lstools_ubuntu`.

Now `MbedLsToolsUbuntu` inherits from `MbedLsToolsLinuxGeneric` (to allow Ubuntu-specific changes in the future). Supposedly since they are identical we can remove the Ubuntu-specific class which adds no functionality right now, but I understand the mbed dev team wants to keep Ubuntu separate.

I tested this on RHEL 6.7 with Freescale FRDM-K64F.

Maybe solves #25.


PrzemekWirkus:
Thank you for your contribution. I will have a look at this shortly!
Did sing [contributor agreement](https://developer.mbed.org/contributor_agreement/) ?

dubek:
hmmm, let me consult my lawyer :-)

PrzemekWirkus:
;)

PS: I can merge it after TechCon (mid of November) most probably because for now we have a feature freeze on this software.

mbed_lstools/lstools_linux_generic.py
```diff
@@ -0,0 +1,311 @@
...
+        """ Find mount points for MBED devices using mount command output
+
+        @return Returns None if mount point not found. Else returns device mount path
+
+        @param dev_name Device name (e.g 'sda')
+        @param mount_list List of all mounted devices (strings from Linux mount shell command)
+
+        @details
+        """
+        mount_media_pattern = "^/[a-zA-Z0-9/]*/" + dev_name  + " on (/[a-zA-Z0-9/]*) "
```
jupe: there should be also `_` in end of detection pattern, like:
`"^/[a-zA-Z0-9/]*/" + dev_name  + " on (/[a-zA-Z0-9_/]*) "`.
This because looks that when multiple devices are connected mount point will be 
* `/media/MBED_` for second device
* `/media/MBED__` for third device
and so on...
* `/media/MBED______`  ... 
.

mbed_lstools/lstools_ubuntu.py
```diff
@@ -15,297 +15,14 @@
...
 
-import re
-import subprocess
+from lstools_linux_generic import MbedLsToolsLinuxGeneric
 
-from lstools_base import MbedLsToolsBase
 
-
-class MbedLsToolsUbuntu(MbedLsToolsBase):
+class MbedLsToolsUbuntu(MbedLsToolsLinuxGeneric):
```
jupe: is this needed at all, because MbedLsToolUbuntu is now direct inheritance without any special functionality ?

dubek:
@PrzemekWirkus I signed the contributor agreement.

mbed_lstools/lstools_ubuntu.py
dubek: The only reason is this comment https://github.com/ARMmbed/mbed-ls/issues/25#issuecomment-151058043 - where @PrzemekWirkus stated that Ubuntu is the main testing platform.

mbed_lstools/lstools_linux_generic.py
dubek: You're right, but that's another issue. Anyway I need to rebase this branch (carefully) because there have been other changes to the code (that I extracted) in the meantime.

dubek:
I pushed this again rebased on top of current master branch.

mbed_lstools/lstools_linux_generic.py
dubek: @jupe - I think you should open a separate PR about your pattern change for multiple devices.

mbed_lstools/lstools_ubuntu.py
jupe: I see

mbed_lstools/lstools_linux_generic.py
jupe: yep, it was in original code too, so this need to be done in separate PR.

jupe: https://github.com/ARMmbed/mbed-ls/pull/33

PR #27, opened 2015-11-09
Adding BBC micro:bit

bridadan:
- Add support for BBC micro:bit


0xc0170:
+1

@PrzemekWirkus shall we fix circleci?

PrzemekWirkus:
Yes, we shall !

PR #31, opened 2015-11-26
Add support for Nuvoton NUC472 NuTiny

ccli8:


ccli8:
Hi,
I'd like to add target ID of Nuvoton NUC472 NuTiny board into lstools_base.py, but meet CircleCI check error which I could not fix. Could you help clarify it?
"1235": "UBLOX_C027",
**"1300": "NUC472-NUTINY",**
"1549": "LPC1549",

PrzemekWirkus:
@ccli8 I think I can't do this. All this mappings are assigned by our partner enablement team. I can only add this if I get a internal info that this is correct mapping to valid name we will be using.
I do not know if you are our partner or not but like I said I can't just add mapping like this. It is a formal process because Target ID mapping is assigned by us (mbed) to proper partners (just like MAC pools are assigned to particular vendors).

ccli8:
@PrzemekWirkus Nuvoton is a partner of ARM mbed. The ID to merge in is assigned by Sam.Grove@arm.com.

PrzemekWirkus:
@sg- Sam, when you are back from holiday please let me know if we can merge it.

sg-:
good to merge!

Issue #41, opened 2016-01-20
Serial number regex not allowing non-hex characters


mvolstad:
The USB serial number search limits the range to 0-9 and A-F. 
Atmel EDBG device serial numbers contains the preample ATML. This is excluded from the search and therefore the COM-port detection does not work. 

Proposal for fix in:
https://github.com/mvolstad/mbed-ls/tree/serialnumber_regex_fix

ciarmcom:
ARM Internal Ref: IOTSYST-775

PrzemekWirkus:
Actually it is not USB serial number but we are searching for ```target id``` of mbed-enabled device. Which is a ASCII hex string. The problem with board you have might be with the way we parse whole USB ID. I need to deeper into that.

PR #43, opened 2016-01-20
Serialnumber regex fix

mvolstad:
Proposed fix for: Serial number regex not allowing non-hex characters #41

### DIFF
--- a/mbed_lstools/lstools_linux_generic.py
+++ b/mbed_lstools/lstools_linux_generic.py
@@ -88,10 +88,10 @@ class MbedLsToolsLinuxGeneric(MbedLsToolsBase):
         self.ERRORLEVEL_FLAG = 0
 
         result = []
-        tidhex = re.compile(r'_([0-9a-fA-F]+)-\d+:\d+')
+        tidpattern = re.compile(r'_([0-9a-zA-Z]+)-\d+:\d+')
         for device in all_devices:
             tid = None
-            m = tidhex.search(device[4])
+            m = tidpattern.search(device[4])
             if m and len(m.groups()):
                 tid = m.group(1)
             mbed = {'mount_point' : device[2],
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ lrwxrwxrwx root  10 Dec  3 09:10 wwn-0x5000cca30ccffb77-part5 -> ../../sda5
 
 ***Note:*** ```mbed-ls``` tools pair only serial ports and mount points (not CMSIS-DAP - yet).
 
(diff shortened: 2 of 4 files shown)


PrzemekWirkus:
@mvolstad Please remove test part from your PR, it is breaking Ci.
Our CI is Linux based (so ```winreg``` module is not found) and have to find another way to unit test this function. I can see change is not so significant so we can merge it when you remove test part.

mvolstad:
@PrzemekWirkus Thanks, I have reverted the tests.

mbed_lstools/lstools_linux_generic.py
```diff
@@ -88,10 +88,10 @@ def list_mbeds(self):
         self.ERRORLEVEL_FLAG = 0
 
         result = []
-        tidhex = re.compile(r'_([0-9a-fA-F]+)-\d+:\d+')
+        tidpattern = re.compile(r'_([0-9a-zA-Z]+)-\d+:\d+')
```
PrzemekWirkus: This is illegal here. We are searching for target id which is specified as hex ascii string. For example:
```
usb-MBED_microcontroller_02400226489a1e6c000000000000000000000000b543e3d4-0:0 -> ../../sdb
```
has a valid ```02400226489a1e6c000000000000000000000000b543e3d4``` Target ID.

In your case ```ATML2127031800007973``` is illegal target id. In any case, does your device have a ```mbed.htm``` file on MSD of interface chip?

PrzemekWirkus: Note: four first characters define [manufacturer_ids](https://github.com/ARMmbed/mbed-ls/blob/master/mbed_lstools/lstools_base.py#L44).
If you wish to detect correctly your devices, e.g. [XPRO_SAM](https://github.com/ARMmbed/mbed-ls/blob/master/mbed_lstools/lstools_base.py#L96) you have to make sure that your target id in USB ID starts with a proper substring, you XPRO case e.g. ```0900``` will be mapped to ```XPRO_SAMR21```.
So any target id like '''0900deadcodedeadcodedeadcode``` will work!

mvolstad:
Yes, we do have an mbed.htm file. This is an example of the redirect URL: http://mbed.org/device/?code=0900030841544d4c32313237303331383030303037393733d10b
Where 0900 is the manufacturer id, and 0308 refers to the version.

The problem is that the part after the version is expected to be the ASCII encoding of the USB serial number ("41544d4c32313237303331383030303037393733" = "ATML2127031800007973"). 

You're right that the USB mass storage class specifies the serial number to be hexadecimal. However, we haven't seen any practical limitations any other places than in mbed-ls. We would prefer to keep the serial number scheme we already have had in place before implementing MSC support. This pull request introduces a reasonable leniency.

PrzemekWirkus:
@mvolstad I did not forget about you :P I just had no time to look closer to it.
In theory we should always look at mbed.htm if USB ID is illegal. And yes, USB ID sometimes is used by manufacturers for different things, no need to bend it to mbed-ls.
Please be patient.

mazimkhan:
I see this change as minor leniency in detecting mbed devices since detection still largely based on vendor Id and ```mbed.htm```. Hence merging!

PR #45, opened 2016-01-26
Add bootloader board ID

c1728p9:
Add the board ID "0000" for bootloader.  This allows boards in this
mode to be seen properly on Linux.

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -42,6 +42,7 @@ class MbedLsToolsBase:
 
     # Dictionary describing mapping between manufacturers' ids and platform name.
     manufacture_ids = {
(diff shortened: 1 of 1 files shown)


0xc0170:
Do all boards have set 0 for bootloader mode?

PrzemekWirkus:
@0xc0170  This is a good question, I do not want to add undefined behavior to mbed-ls.

PrzemekWirkus:
Looks like our Linux implementation is more strict and adding 0000 is not a good idea. Changing underlying Linux implementation is.

c1728p9:
Thanks for looking into this.  I'll close this PR and look at the Linux implementation instead.  Any suggestions as to where to look this make this improvement would be appreciated.

Issue #44, opened 2016-01-26
Boards in bootloader mode show up on Windows but not Linux


c1728p9:
Running mbedls on Linux does not show any boards in bootloader mode.  Windows shows connected boards in bootloader mode even though they are unknown:
```
+---------------------+----------------------------+-------------------+-------------------+--------------------------------------------------------+
|platform_name        |platform_name_unique        |mount_point        |serial_port        |target_id                                               |
+---------------------+----------------------------+-------------------+-------------------+--------------------------------------------------------+
|unknown              |None[0]                     |D:                 |unknown            |0000000026334e450053400f5e0e00134eb10000646c0000        |
+---------------------+----------------------------+-------------------+-------------------+--------------------------------------------------------+
```

Is this a bug in the Linux implementation?  Would it be possible to add support for boards in bootloader mode?  This would be very useful for DAPLink testing.

c1728p9:
Pushed up #45 which fixes this on Linux.  Not sure if this is the appropriate fix, but it works for me.  Let me know what you think.

ciarmcom:
ARM Internal Ref: IOTSYST-798

PrzemekWirkus:
This is not a bug, we just do not support boards in bootloader mode at all in mbed-ls.
But I can see you wanna add this feature... so this is a feature request. I will check your PR and if it is OK I will merge, if not I will let you know how it looks like.

c1728p9:
I put up PR #46 to add support for this on Linux.

PrzemekWirkus:
Released with v0.1.27

PR #46, opened 2016-01-26
Linux: Add support for boards without serial port

c1728p9:
On Linux allow orphan boards without a serial port to be displayed
rather than filtering them out.  This allows boards in bootloader mode
to be detected.


PrzemekWirkus:
This PR actually makes sense :P

PrzemekWirkus:
+1

PrzemekWirkus:
This is a response for https://github.com/ARMmbed/mbed-ls/issues/44

PR #47, opened 2016-01-26
Windows: add support for Serial on ARM SSG Devices

fvincenzo:
This patch adds support for Serial on ARM SSG Devices.

Signed-off-by: Vincenzo Frascino <vincenzo.frascino@arm.com>


PrzemekWirkus:
@fvincenzo Vincenzo, what is the USB ID of the device you want to add support to?

fvincenzo:
After further debugging I found a way to enable the serial port in the standard way.

Thanks,
Vincenzo

PR #51, opened 2016-02-16
Added support for OS X El Capitan

jjones646:


jjones646:
I've made some changes for getting El Capitan on board. I realized today that it currently will not work through hubs, so I've updated it to print out some debugging info at the moment and am looking into that.

I can currently only test this with just El Capitan and Yosemite.

PrzemekWirkus:
Can you provide some more description for your PR??
And is it possible to squash all 51 commits to one / two ? ;)

PrzemekWirkus:
@jjones646 You probably want to add El Capitan detection [here](https://github.com/ARMmbed/mbed-ls/blob/master/mbed_lstools/main.py#L48) and prepare new file with proper OS port implementation:
```python
from lstools_base import MbedLsToolsBase

class MbedLsToolsElCapitan(MbedLsToolsBase):
... # El Capitan implementation
```

jjones646:
I believe it's working now. It runs the `ioreg -a -t -l -r -c IOSerialBSDClient` command to get a listing of the full paths from the root for all connected serial devices. It then uses the serial number from there to match up drives that share the same serial number from the `system_profiler SPUSBDataType -detailLevel full -xml` command. All of that is recursively parsed from the root level of both trees.

@PrzemekWirkus I'll compress down the commits. As you can see, I had a bit of trouble doing that so far. Also, I assuming you're talking about the `usb_controllers = ['AppleUSBXHCI', 'AppleUSBUHCI', 'AppleUSBEHCI', 'AppleUSBOHCI']` part when you say proper os port implementation? I'm no expert with any of this stuff and have only used and tested this with multiple LPC1768 mbeds. So would that still be necessary with the approach described above?

jjones646:
This should be ready now. I added pretty verbose output showing exactly what gets selected when you pass `-d` flag using the cli.

c1728p9:
I get an exception when testing this on Yosemite
```
  ...
     mbed_list = lstools.list_mbeds()
  File "/Users/arm/DAPLink-test_hw/virtual_env/lib/python2.7/site-packages/mbed_lstools/lstools_darwin.py", line 56, in list_mbeds
    volumes = self.get_mbed_volumes()
  File "/Users/arm/DAPLink-test_hw/virtual_env/lib/python2.7/site-packages/mbed_lstools/lstools_darwin.py", line 224, in get_mbed_volumes
    self.treeShow.show(serial_devs)
  File "/Users/arm/DAPLink-test_hw/virtual_env/lib/python2.7/site-packages/mbed_lstools/term_formats.py", line 86, in show
    (h, w) = getTermDims()
ValueError: need more than 0 values to unpack
```

jjones646:
@c1728p9 Is it working now?

c1728p9:
I don't get a crash anymore but mbed-ls no longer picks up some the boards I have attached.  Could you try rebasing your PR on top of master?

jjones646:
I've rebased on top of master. Also added another usb key that gets searched for supporting targets that may be running older firmware.

c1728p9:
I gave the re-based patch a try but it still doesn't pick up on some of the boards.  Below are logs from mbed-ls on pypi and from your patch.
[mbedls.txt](https://github.com/ARMmbed/mbed-ls/files/174846/mbedls.txt)
[mbedls_new.txt](https://github.com/ARMmbed/mbed-ls/files/174847/mbedls_new.txt)

Issue #53, opened 2016-02-24
daplink version is not shown when calling mbedls


jupe:
When calling mbedls without parameters it show `unknown` as daplink_version, json format seems to work correctly. Tested in ubuntu 14.04 VM.
```
jusvat01@ubuntu:~/git/github/ARMmbed/raas-daemon$ mbedls 
+---------------+----------------------+-------------------------+--------------+--------------------------------------------------+-----------------+
| platform_name | platform_name_unique | mount_point             | serial_port  | target_id                                        | daplink_version |
+---------------+----------------------+-------------------------+--------------+--------------------------------------------------+-----------------+
| K64F          | K64F[0]              | /media/jusvat01/DAPLINK | /dev/ttyACM0 | 0240000028884e45002500106bf000448021000097969900 | unknown         **|**
```
json format show correctly
```
jusvat01@ubuntu:~/git/github/ARMmbed/raas-daemon$ mbedls --json
[
    {
        "daplink_#_daplink_firmware_-_see_https": "//mbed.com/daplink", 
        "daplink_auto_reset": "0", 
        "daplink_automation_allowed": "0", 
        "daplink_daplink_mode": "Interface", 
        "daplink_git_sha": "c765cbb590f57598756683254ca38b211693ae5e", 
        "daplink_hif_id": "97969900", 
        "daplink_interface_crc": "0x26764ebf", 
        "daplink_interface_version": "0240", 
        "daplink_local_mods": "0", 
        "daplink_unique_id": "0240000028884e45002500106bf000448021000097969900", 
        "daplink_usb_interfaces": "MSD, CDC, HID", 
        "mount_point": "/media/jusvat01/DAPLINK", 
        "platform_name": "K64F", 
        "platform_name_unique": "K64F[0]", 
        "serial_port": "/dev/ttyACM0", 
        "target_id": "0240000028884e45002500106bf000448021000097969900", 
        "target_id_mbed_htm": "0240000028884e45002500106bf000448021000097969900", 
        "target_id_usb_id": "0240000028884e45002500106bf000448021000097969900"
    }
]
```

jupe:
looks that this related some how to custom daplink firmware.

PrzemekWirkus:
Please copy-paste DETAILS.TXT here and MBED.HTM for debug :)

PrzemekWirkus:
mbed.html:
```html
<!doctype html>
<!-- mbed Platform Website and Authentication Shortcut -->
<html>
<head>
[...]

PR #54, opened 2016-02-25
Add support for DETAILS.TXT and mbed.htm for DAPlink >=0240

PrzemekWirkus:
This is response for [daplink version is not shown when calling mbedls](https://github.com/ARMmbed/mbed-ls/issues/53).


jupe:
lgtm

PR #64, opened 2016-03-29
Add support for Python 3

PrzemekWirkus:
Changes:
* Small changes to make ```mbed-ls``` usable with Python 3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,7 @@ lrwxrwxrwx root  10 Dec  3 09:10 wwn-0x5000cca30ccffb77-part5 -> ../../sda5
 
 ***Note:*** ```mbed-ls``` tools pair only serial ports and mount points (not CMSIS-DAP - yet).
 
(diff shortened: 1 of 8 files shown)


PrzemekWirkus:
This one is first thing to fix here ;)
```bash
$ python setup.py install

Traceback (most recent call last):
  File "setup.py", line 38, in <module>
    long_description=read('README.md'),
  File "setup.py", line 33, in read
    return open(os.path.join(os.path.dirname(__file__), fname), encoding='utf8').read()

python setup.py install returned exit code 1

TypeError: 'encoding' is an invalid keyword argument for this function Action failed: python setup.py install
```

jupe:
you should add py3 support for circle conf file as well so that ci verify both python versions..

PrzemekWirkus:
@jupe I will do it when this port is ready, still no time to finish it.

PR #67, opened 2016-03-31
Fix serial port recognition in ubuntu 15.10

mjs-arm:
The serial port recognition code pattern matches against a string
prefixed with usb- however on ubuntu 15.10 these strings are prefixed
with pci-.  This patch relaxes the pattern to allow and 3 character
prefix.


PrzemekWirkus:
@mjs-arm Thanks for your contribution.
Please:
1. Add to description example of string with ```pci-```...
2. Please update unit tests [here](https://github.com/ARMmbed/mbed-ls/blob/master/test/os_linux_generic.py#L289) and [here](https://github.com/ARMmbed/mbed-ls/blob/master/test/os_linux_generic.py#L301).
3. Run ```python setup.py test``` to check for regressions. You can paste output to PR description.

PrzemekWirkus:
You've replaced this regex:
```python
"(usb-[0-9a-zA-Z_-]*_[0-9a-zA-Z]*-.*$)"
```
with
```python
self.name_link_pattern= "(...-[0-9a-zA-Z_-]*_[0-9a-zA-Z]*-.*$)"
```

That is too generic pattern IMHO. I would still narrow down search to currently available patterns: ```usb-``` and ```pci-```.

Maybe something like this:
```python
self.name_link_pattern = '((%s)-[0-9a-zA-Z_-]*_[0-9a-zA-Z]*-.*$)'% ('|'.join(["pci", "usb"]))
```

PrzemekWirkus:
I'm gonna do this here: https://github.com/ARMmbed/mbed-ls/pull/68
Thanks for your help Marcus!

Issue #75, opened 2016-05-05
Add ID for Unlisted boards


BlackstoneEngineering:
## Problem
There are going to be boards that use DAPLink but are not part of the mbed-enabled program, these boards should have a known 'unknown' identifier that is used so they do not gank up the system. 

## Solution
Add a known 'Unknown Board' ID, I suggest using `OSHW`. 
https://github.com/ARMmbed/mbed-ls/blob/master/mbed_lstools/lstools_base.py

ciarmcom:
ARM Internal Ref: IOTSFW-2521

PrzemekWirkus:
@BlackstoneEngineering Can you provide before and after mbedls printouts please?

BlackstoneEngineering:
Closing this issue, see https://github.com/ARMmbed/mbed-ls/pull/76

PR #76, opened 2016-05-05
added board ID for unknown daplink implementations

BlackstoneEngineering:
Fix for https://github.com/ARMmbed/mbed-ls/issues/75

@bridadan

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -176,6 +176,7 @@ class MbedLsToolsBase:
         "9012": "SEEED_TINY_BLE",
         "9900": "NRF51_MICROBIT",
         "FFFF": "K20 BOOTLOADER",
+        "OSHW":"Unknown DAPLink Implementation",
         "RIOT": "RIOT",
     }
 


mbed_lstools/lstools_base.py
```diff
@@ -176,6 +176,7 @@ def __init__(self):
         "9012": "SEEED_TINY_BLE",
         "9900": "NRF51_MICROBIT",
         "FFFF": "K20 BOOTLOADER",
+        "OSHW":"Unknown DAPLink Implementation",
```
bridadan: How about just `UNKNOWN_TARGET` instead of `Unknown DAPLink Implementation`? Or `OSHW`?

mbed-ls doesn't list "DAPLink implementations", it uses the DAPLink implementation to ID what target is on the other side of the DAPLink circuit.

PrzemekWirkus: +1 for `UNKNOWN_TARGET` or similar.

BlackstoneEngineering: `UNKNOWN_DAPLINK` would work. Its not an unknown target, its a known identifier to a group of Daplink boards that do not have mbed platform support.

bridadan: I guess I'm nitpicking now at terminology :) mbed-ls doesn't list Daplink boards, it lists mbed enabled platforms, just like you said. It doesn't require DAPLink.

If not `UNKNOWN_TARGET`, how about `UNKNOWN_PLATFORM` or `COMMUNITY_PLATFORM`? I kind of like `COMMUNITY_PLATFORM` since the manufacturer id is actually "known". It's a general ID that can be used by anyone. That way we're staying consistent with the `platform` terminology used throughout the mbed-ls code base and the command line results. (https://github.com/ARMmbed/mbed-ls/blob/5587839710d02f91138d065a187f3ab1665ecc9e/mbed_lstools/lstools_base.py#L418)

PrzemekWirkus:
@BlackstoneEngineering Do we have any solid requirements or some `mbed-enabled` requirements (?) to support this PR? How the whole mechanism works? Non-partners (Open Source community) will be notified to use `OSHW` in their HW implementations.

Currently each user can add custom naming for target detectable with `mbed-ls`. See [here](https://github.com/ARMmbed/mbed-ls#mocking-new-or-existing-target-to-custom-platform-name). 
* With this simple feature everyone can distribute HW with simple mapping file and makes sure `mbed-ls` grabs his platform. Note that this setting is not global, but per directory where we call `mbed-ls`.

mbed_lstools/lstools_base.py
BlackstoneEngineering: The ID's you are using to check if it is an mbed board is the daplink id, there is no difference. It comes straight from @sg- , use `UNKNOWN_DAPLINK` or similar. If you want to argue it go talk to him.

BlackstoneEngineering:
@PrzemekWirkus yes, this is a direct requirement from mbed enabled. For platforms that want to use daplink but dont want to go through the mbed enabled process we need a known daplink ID number that Identifies them as such. The reason this is being added to mbed-ls, which normally wouldnt care, is that the list of ID's for mbed-ls is often cited as the go to list for a publically available list of board ID's, hence why we need to have the `OSHW` id included in mbed-ls. 

Please just go ahead and merge this, it doesnt make your life harder but it does make our lives in Dev Enablement much easier.

sg-:
OSHW isnt a good name. for an unknown board use something else.

BlackstoneEngineering:
@sg- we talked about this the other day, you already approved oshw.

BlackstoneEngineering:
K, since this is a preventative measure I will re-open this when it becomes an issue. Was trying to anticipate a problem, but this is too much of a hassle right now, closing and will re-open when pertinant.

PR #77, opened 2016-05-06
Add command execution feature with subprocess.communicate()

PrzemekWirkus:
## Description

When enumerating for Ubuntu calling ```mount | grep vfat``` hangs when there are many mount points. With some mount systems like LDM we can have many mount points listed with mount command. If so previous ```Popen.wait()```solution .

When a pipe's buffer fills up (typically 4KB or so), the writing process stops until a reading process has read some of the data in question; but here you're reading nothing until the subprocess is done, hence the deadlock.

Example community discussion regarding wait() vs. communicate() use [here](http://stackoverflow.com/questions/1445627/how-can-i-find-out-why-subprocess-popen-wait-waits-forever-if-stdout-pipe).

## Changes:
* Use communicate() instead of wait() when calling process with Popen(). 
* Avoid wait forever for >4kB outputs.

## Example for wait() vs communicate()
```python
# bad code
p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
p.wait()  # May block if stdout > 4kB

# good code
p = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
_stdout, _stderr = p.communicate() # Will read all data from stdout/stderr
```

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -577,6 +577,7 @@ class MbedLsToolsBase:
                 return result
         return None
 
+    @staticmethod
     def run_cli_process(cmd, shell=True):
         """! Runs command as a process and return stdout, stderr and ret code
         @param cmd Command to execute
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -576,3 +576,14 @@ class MbedLsToolsBase:
                     self.debug(self.scan_html_line_for_target_id.__name__, (m.groups(), result))
                 return result
         return None
+
+    def run_cli_process(cmd, shell=True):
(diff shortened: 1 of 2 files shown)


PrzemekWirkus:
Tested on RPI2:
```
| K64F          | K64F[0]              | /media/usb0 | /dev/ttyACM0 | 0240000029304e450008500878a2003a1951000097969900 | 0241            |
| K64F          | K64F[1]              | /media/usb1 | /dev/ttyACM1 | 0240000033514e450054500585d4001de981000097969900 | 0241            |
| K64F          | K64F[2]              | /media/usb2 | /dev/ttyACM2 | 0240000030514e45002620067d7e001d1f91000097969900 | 0241            |
```

PR #86, opened 2016-06-14
[Silicon Labs] Added Pearl Gecko to mbed-ls

stevew817:
Pearl Gecko (EFM32PG) was missing from the mbed-ls board list.


PrzemekWirkus:
@sg- FYI

Issue #100, opened 2016-06-22
Mocked platforms timeout on tests


BlackstoneEngineering:
## Problem
I tried to mock the hexiwear device by doing `mbedls --mock=0214:HEXIWEAR`, this allowed me to then run `singletest.py -j 0 --tc GCC_ARM -V`, but all tests time out. I then instead went and modified mbedls, adding the appropriate line to mbedls to enable hexiwear, re-installed, deleted the build directory, and re-ran the tests. Then all tests worked. 

## Conclusion
Something in the `mock` subcommand doesnt work with tests. Not sure what

## Solution
make mock work with testing infrastructure as this is the primary way we expect users to test their platforms before submitting a PR.

BlackstoneEngineering:
Testing on mbedmicro/mbed

ciarmcom:
ARM Internal Ref: IOTSFW-2740

PrzemekWirkus:
@BlackstoneEngineering Austin did you take under consideration that `--mock` works only in the same directory not globally. Description is in [this](https://github.com/ARMmbed/mbed-ls#mocking-example-with-freescale-k64f-platform) chapter in the description under third block.

I quote: **"We can observe changes immediately. Please note this change only works in the same directory because we save .mbedls-mock file locally:"**.

Please try with pre HEX mbed-ls and this workflow:
```
$ cd mbed
$ cd workspace_tools
$ mbedls --mock=0214:HEXIWEAR
$ singletest.py --auto --tc GCC_ARM -V
```

I was unsure if command you've put here (`singletest.py -j 0 --tc GCC_ARM -V) is exactly what you were using.

I'm assuming that you were actually using `--auto` to detect HEX platform and forced `GCC_ARM` toolchain using `--tc` command line switch.

BlackstoneEngineering:
Yes, i took the local folder into consideration. I even had @bridadan double check to make sure I wasn't crazy. Using `--mock` worked in so far as identifying the board correctly, but did not work when trying to run the test suite (singletest.py), however when I modified mbed-ls-git and reinstalled the test suite worked correctly.

This is going to be a terrible usability bug when we try to get people in the general populace / partners to port their stuff with minimal support.

PrzemekWirkus:
Done via #106

BlackstoneEngineering:
Awesome, closing issue

Issue #99, opened 2016-06-22
mock needs to be global


BlackstoneEngineering:
## Problem
When using the `--mock` subcommand it creates a folder in what ever directory I am in, I then must use that mocked target from within that folder. This use case makes sense as a backend tool but not as a user facing tool.

## Solution
When adding a 'mocked` device it should be added globally by default with the option to be folder specific for when needed on the backend. This reflects the transition of this tool from a backend tool to a frontend user facing tool. 

## Common Use Case
A user wants to test their new platform, to do this they must run the tests from the mbedmicro/mbed/tools directory. All of these tools require the board to be usable from mbedls. Thus the user runs `mbedls --mock=xxxx:PlatformName` and then the user can run the tests and validate their port. Under the current model the user would have to run all the tests from the base folder where they start the mock, but if you're doing active SDK development then this doesnt make sense as your prompt will be all over the place. Makes much more sense to add it globally so the mocked platform can be used by any tool / test anywhere on the local system.

ciarmcom:
ARM Internal Ref: IOTSFW-2738

PrzemekWirkus:
Mock can be global only per OS user. Do you think this would be enough?

BlackstoneEngineering:
Yes, that should be good enough, for the typical users (windows, single user laptop/desktop machine) this should net the same effect.

PrzemekWirkus:
Done via #106

BlackstoneEngineering:
cool, closing issue

PR #106, opened 2016-07-04
Add global $HOME/.mbed-ls/.mbedls-mock storage for mock file

PrzemekWirkus:
Changes:
* Add global mock file/directory `$HOME/.mbed-ls/.mbedls-mock` storage for mock file
* For backward compatibility we will for now read first local .mbedls-mock file and skip one in $HOME
* Addressing: *mock needs to be global*, see #99 and *Mocked platforms timeout on tests*, see #100


PrzemekWirkus:
@bridadan @BlackstoneEngineering Please review and give comments.

PR #111, opened 2016-07-19
Replace lockfile context mgr usage with lock.aquire() and lock.release()

PrzemekWirkus:
# Description

On RP2 Linux we have a deadlock when using root account so we may want to drop mocking feature but have non-blocking mbed-ls with timeout=.5 sec on file lock acquire.

# Rationale

To catch LockTImeout exception we can't use context mgr. We need to use explicit
calls to `acquire(timeout=1)` and `release()` to catch timeout.
Timeout may occur on some Resource Managers instance.

LockFile objects support the context manager protocol used by the statement:with
statement. The timeout option is not supported when used in this fashion. While
support for timeouts could be implemented, there is no support for handling the
eventual Timeout exceptions raised by the __enter__() method, so you would have
to protect the with statement with a try statement. The resulting construct
would not be any simpler than just using a try statement in the first place.

# Example timeout
```
mbedls
error: Timeout waiting to acquire lock for /home/pi/.mbed-ls/mbedls-lock
| NRF51_DK      | NRF51_DK[0]          | /mnt/MBED__ | /dev/ttyACM2 | 1100021944203120324C46383030343239303038B9C6DFD8 | 0219            |
| NRF51_DK      | NRF51_DK[1]          | /mnt/MBED_  | /dev/ttyACM1 | 1100021944203120324C46383130383131323033B9C6DFD8 | 0219            |
```

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -22,7 +22,7 @@ import json
 import lockfile
 from os import listdir
 from os.path import isfile, join
-from lockfile import LockFailed
+from lockfile import LockFailed, LockTimeout
 
 class MbedLsToolsBase:
     """ Base class for mbed-lstools, defines mbed-ls tools interface for mbed-enabled devices detection for various hosts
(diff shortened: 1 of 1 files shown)


PrzemekWirkus:
@bridadan @adbridge @mazimkhan FYI

Issue #137, opened 2016-10-04
Latest ST-Link fails to list serial port on Windows


bridadan:
Using the latest version of ST-Link (version 2.28.16 found here: http://www.st.com/content/st_com/en/products/embedded-software/development-tool-software/stsw-link007.html), my nucleo devices do not have a serial port that is picked up be mbedls (though I can see it in Device Manager)

Using Windows 7 64 bit and mbed-ls version 1.2.5.

**NOTE:** This is not an issue in mbed-ls version 1.2.4.

```
debug @MbedLsToolsWin7.mbed_lstools.main: mbed-ls ver. 1.2.5
debug @MbedLsToolsWin7.mbed_lstools.main: host: ('nt', 'Windows', '7', '6.1.7601', 'win32')
debug @MbedLsToolsWin7.get_mbed_devices: ('\\DosDevices\\E:', '_??_USBSTOR#Disk&Ven_MBED&Prod_microcontroller&Rev_1.0#a&19934e84&0&066BFF494951785087022947&0#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}')
debug @MbedLsToolsWin7.get_mbed_devices: ('\\DosDevices\\F:', '_??_USBSTOR#Disk&Ven_MBED&Prod_microcontroller&Rev_1.0#8&4bb3ea4&0&0669FF515456707067164717&0#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}')
debug @MbedLsToolsWin7.get_mbed_devices: ('\\DosDevices\\G:', '_??_USBSTOR#Disk&Ven_MBED&Prod_microcontroller&Rev_1.0#a&1a08f641&0&0670FF494951785087022432&0#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}')
debug @MbedLsToolsWin7.get_mbed_devices: ('\\DosDevices\\H:', '_??_USBSTOR#Disk&Ven_MBED&Prod_microcontroller&Rev_1.0#a&1d9f9a1d&0&066EFF485649785087174252&0#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}')
debug @MbedLsToolsWin7.get_mbeds: ('E:', 'microcontroller')
debug @MbedLsToolsWin7.get_mbeds: ('F:', 'microcontroller')
debug @MbedLsToolsWin7.get_mbeds: ('G:', 'microcontroller')
[...]

PR #138, opened 2016-10-04
Fixing tid detection in Windows

bridadan:
This fixes an issue with detecting NUCLEO boards after 316195e4d96cb47cb359618849b70c713c317a7e on Windows. It makes use of the '&' and '#' delimeter in the USB string from the registry to find the
correct tid.

Note: This is high priority as NUCLEO detection on Windows is currently broken as of v1.2.5 as detailed by #137 

FYI @adbridge 
Please take a look @mazimkhan

### DIFF
--- a/mbed_lstools/lstools_win7.py
+++ b/mbed_lstools/lstools_win7.py
@@ -161,7 +161,7 @@ class MbedLsToolsWin7(MbedLsToolsBase):
         for mbed in self.get_mbed_devices():
             mountpoint = re.match('.*\\\\(.:)$', mbed[0]).group(1)
             # TargetID is a hex string with 10-48 chars
-            tid = re.search('[0-9A-Za-z]{10,48}', mbed[1]).group(0)
+            tid = re.search('[&#]([0-9A-Za-z]{10,48})[&#]', mbed[1]).group(1)
             mbeds += [(mountpoint, tid)]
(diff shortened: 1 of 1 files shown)


bridadan:
The reason why this wasn't an issue before is apparent from this debug log line:

```
debug @MbedLsToolsWin7.get_mbed_devices: ('\\DosDevices\\G:', '_??_USBSTOR#Disk&Ven_MBED&Prod_microcontroller&Rev_1.0#a&1a08f641&0&0670FF494951785087022432&0#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}')
```

The regex I modified is used against this string in particular:

```
_??_USBSTOR#Disk&Ven_MBED&Prod_microcontroller&Rev_1.0#a&1a08f641&0&0670FF494951785087022432&0#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}
```

The regex before 316195e4d96cb47cb359618849b70c713c317a7e matched the following two segments: `0670FF494951785087022432` and `00a0c91efb8b`. And mbed-ls just picked the first one.

The regex after 316195e4d96cb47cb359618849b70c713c317a7e matched the following three segments: `microcontroller`, `0670FF494951785087022432`, `00a0c91efb8b`. mbed-ls picked the first one again, which is not a valid target id.

The regex after my PR now matches only `0670FF494951785087022432`, which is then used for the target id.

Issue #130, opened 2016-09-13
Move 'lockfile' dependency to 'fasteners' or 'oslo.concurrency'


bridadan:
As mentioned at the top of the [lockfile documentation](https://pypi.python.org/pypi/lockfile/0.12.2), the package is now deprecated. It also a conflict with a default Ubuntu pip package as described here: https://github.com/ARMmbed/mbed-ls/issues/128#issuecomment-246587081

We can probably take care of both issues by moving to one of the recommended packages from the lockfile docs: [fasteners](https://pypi.python.org/pypi/fasteners) or [oslo.concurrency](http://docs.openstack.org/developer/oslo.concurrency/).

mazimkhan:
@bridadan are you working  on it?

bridadan:
@mazimkhan Not currently :smile:

adbridge:
ARM Internal Ref: IOTSFW-2924

mazimkhan:
Fixed by https://github.com/ARMmbed/mbed-ls/pull/139

bridadan:
Thanks @mazimkhan !

PR #139, opened 2016-10-06
Replaced use of lockfile with fasteners and optimised locking code.

mazimkhan:
Changes:
- Replaced ```lockfile``` with ```fasteners``` 
- Created a decorator for methods that require inter process locking.

@bridadan @adbridge Please review.

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -17,12 +17,44 @@ limitations under the License.
 
 import re
 import os
+import sys
 from os.path import expanduser
 import json
-import lockfile
 from os import listdir
 from os.path import isfile, join
-from lockfile import LockFailed, LockTimeout
+from fasteners import InterProcessLock
+
+
+def timed_mbedls_lock(timeout):
+    """
+    Implements Inter Process Lock using fasteners.InterProcessLock and adding a timeout feature to avoid wait forever.
+
+    :param timeout:     Time to wait for acquiring lock. Else an exception is raised.
+    :return:
+    """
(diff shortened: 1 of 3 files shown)


mbed_lstools/lstools_base.py
```diff
@@ -260,13 +285,13 @@ def list_manufacture_ids(self):
 
         return pt.get_string()
 
+    @timed_mbedls_lock(60)
```
bridadan: Digging the decorator syntax, nice!

mbed_lstools/lstools_base.py
```diff
@@ -323,20 +337,8 @@ def write_mock_file(filename, mock_ids):
...
-        while not lock.i_am_locking():
-            try:
-                lock.acquire(timeout=1)
-            except LockTimeout as e:
-                lock.break_lock()
-                lock.acquire()
-                self.debug(self.mock_read.__name__, str(e))
-        self.debug(self.mock_write.__name__, "locked '%s'"% lock.path)
-
         result = write_mock_file(self.MOCK_HOME_FILE_NAME, mock_ids)
```
bridadan: Any reason why the function `write_mock_file` can't be inlined here? I suppose this was mostly keep it parallel to how `mock_read` is written?

mazimkhan:
Fixes issue #130

mbed_lstools/lstools_base.py
bridadan: I'm fine with revisiting this in another PR if we decide to do so, shouldn't block this PR!

Issue #141, opened 2016-10-17
v1.2.8 doesn't work on windows 7


jupe:
Commit 9a3121e003e5a436f48f6b62a17373d3deb148f1

```
c:\github\ARMmbed\mbed-ls>python setup.py install
...
c:\github\ARMmbed\mbed-ls>mbedls
Traceback (most recent call last):
  File "C:\Python27\Scripts\mbedls-script.py", line 9, in <module>
    load_entry_point('mbed-ls==1.2.8', 'console_scripts', 'mbedls')()
  File "build\bdist.win32\egg\mbed_lstools\main.py", line 217, in mbedls_main
  File "build\bdist.win32\egg\mbed_lstools\lstools_base.py", line 560, in get_string
  File "build\bdist.win32\egg\mbed_lstools\lstools_base.py", line 425, in list_mbeds_ext
  File "build\bdist.win32\egg\mbed_lstools\lstools_win7.py", line 48, in list_mbeds
  File "build\bdist.win32\egg\mbed_lstools\lstools_win7.py", line 69, in discover_connected_mbeds
  File "build\bdist.win32\egg\mbed_lstools\lstools_win7.py", line 141, in get_connected_mbeds
  File "build\bdist.win32\egg\mbed_lstools\lstools_win7.py", line 164, in get_mbeds
AttributeError: 'NoneType' object has no attribute 'group'

c:\github\ARMmbed\mbed-ls>git rev-parse HEAD
9a3121e003e5a436f48f6b62a17373d3deb148f1
```

Seems that same happens also when installing via pip registery.

jupe:
seems that this patch fix my issue:
```
diff --git a/mbed_lstools/lstools_win7.py b/mbed_lstools/lstools_win7.py
index 675f8ec..ae5fd45 100644
--- a/mbed_lstools/lstools_win7.py
+++ b/mbed_lstools/lstools_win7.py
@@ -161,7 +161,9 @@ class MbedLsToolsWin7(MbedLsToolsBase):
         for mbed in self.get_mbed_devices():
             mountpoint = re.match('.*\\\\(.:)$', mbed[0]).group(1)
             # TargetID is a hex string with 10-48 chars
-            tid = re.search('[&#]([0-9A-Za-z]{10,48})[&#]', mbed[1]).group(1)
+            m = re.search('[&#]([0-9A-Za-z]{10,48})[&#]', mbed[1])
+            if not m: continue
+            tid = m.group(1)
```

[...]

PR #143, opened 2016-10-28
Fix issue were a method is attempted to be called on a NoneType.

adbridge:
In get_mbeds() if the regular expression for looking for a tid fails then
a NoneType occurs. The code then attempts to call the group() function on
this which causes an exception.
This fix adds a check for the regular expression returning a NoneType
before trying to call the method.


adbridge:
@bridadan or @mazimkhan please review

adbridge:
Fixes #141

PR #146, opened 2016-11-07
Fix retarget handling

stevew817:
Retargeting was only loosely supported based on the code using mbed-lstools having to specifically call retarget() on the lstools to 'get' the retargeted platform.
With this fix the retarget functionality is always implemented (as long as the retarget spec file is found), leading to a much more consistent behaviour.

This is a fix for https://github.com/ARMmbed/greentea/issues/202

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -74,7 +74,7 @@ class MbedLsToolsBase:
                                       MbedLsToolsBase.MBEDLS_HOME_DIR, MbedLsToolsBase.MBEDLS_GLOBAL_LOCK)
         self.mbedls_get_mocks()
         # Make sure retargeting is applied if retarget file exists
-        self.retarget()
+        self.retarget(True)
 
     # Which OSs are supported by this module
     # Note: more than one OS can be supported by mbed-lstools_* module
@@ -368,12 +368,17 @@ class MbedLsToolsBase:
                     str(e)))
         return {}
 
-    def retarget(self):
+    def retarget(self, enable=True):
         """! Enable retargeting
         @details Read data from local retarget configuration file
-        @return Retarget data structure read from configuration file
+        @param enable True  Try to read data from local retarget configuration file
+                      False Remove all retarget data read (disabling retarget functionality)
+        @return Retarget data structure read from configuration file, empty if disabled or no data found
         """
-        self.retarget_data = self.retarget_read()
+        if enable:
+            self.retarget_data = self.retarget_read()
+        else:
+            self.retarget_data = {}
         return self.retarget_data
 
     def mock_manufacture_ids(self, mid, platform_name, oper='+'):
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -73,6 +73,8 @@ class MbedLsToolsBase:
         self.lock_file = os.path.join(MbedLsToolsBase.HOME_DIR,
                                       MbedLsToolsBase.MBEDLS_HOME_DIR, MbedLsToolsBase.MBEDLS_GLOBAL_LOCK)
         self.mbedls_get_mocks()
+        # Make sure retargeting is applied if retarget file exists
+        self.retarget()
 
     # Which OSs are supported by this module
     # Note: more than one OS can be supported by mbed-lstools_* module
--- a/mbed_lstools/main.py
+++ b/mbed_lstools/main.py
@@ -170,8 +170,8 @@ def mbedls_main():
     mbeds.debug(__name__, "mbed-ls ver. " + get_mbedls_version())
     mbeds.debug(__name__, "host: " +  str((mbed_lstools_os_info())))
 
-    if not opts.skip_retarget:
-        mbeds.retarget()
+    if opts.skip_retarget:
+        mbeds.retarget(False)
 
     if opts.list_platforms:
         print mbeds.list_manufacture_ids()


mbed_lstools/lstools_base.py
```diff
@@ -73,6 +73,8 @@ def __init__(self):
         self.lock_file = os.path.join(MbedLsToolsBase.HOME_DIR,
                                       MbedLsToolsBase.MBEDLS_HOME_DIR, MbedLsToolsBase.MBEDLS_GLOBAL_LOCK)
         self.mbedls_get_mocks()
+        # Make sure retargeting is applied if retarget file exists
+        self.retarget()
```
mazimkhan: There is this command line option ```--skip-retarget``` to disable re-target when the file may be present. 
I would suggest putting that check inside method ```retarget()```.

stevew817:
@mazimkhan @bridadan I've updated the retarget() function to take the skip-retarget flag into account when called from the commandline, and allow programs interfacing directly to disable retargeting (enabled by default).

mbed_lstools/lstools_base.py
```diff
@@ -366,12 +368,17 @@ def retarget_read(self):
                     str(e)))
         return {}
 
-    def retarget(self):
+    def retarget(self, enable=True):
```
bridadan: It seems funny to have a function that can optionally do nothing. Wouldn't you instead just want to optionally call the function?

mbed_lstools/lstools_base.py
```diff
@@ -73,6 +73,8 @@ def __init__(self):
         self.lock_file = os.path.join(MbedLsToolsBase.HOME_DIR,
                                       MbedLsToolsBase.MBEDLS_HOME_DIR, MbedLsToolsBase.MBEDLS_GLOBAL_LOCK)
         self.mbedls_get_mocks()
+        # Make sure retargeting is applied if retarget file exists
+        self.retarget(True)
```
bridadan: When you're using the API directly, this implementation has the downside of not being able to disable retargetting.

I'd suggest instead of passing in a parameter to retarget, add an optional parameter to the constructor of this class that enables retarget by default. Something like `def __init__(self, retarget=True):` Then you can just conditionally call `retarget` based on this variable.

mbed_lstools/main.py
```diff
@@ -170,8 +170,8 @@ def get_mbedls_version():
     mbeds.debug(__name__, "mbed-ls ver. " + get_mbedls_version())
     mbeds.debug(__name__, "host: " +  str((mbed_lstools_os_info())))
 
-    if not opts.skip_retarget:
-        mbeds.retarget()
+    if opts.skip_retarget:
+        mbeds.retarget(False)
```
bridadan: Based on my proposed changes above, you'd instead pass a parameter to the `create` function above that enables/disables retargeting. That way the appropriate parameter can be passed to the constructor of the `MbedLsTools*` class.

mbed_lstools/lstools_base.py
stevew817: Well, yes, it does have the ability to disable retargeting: you remove your mbedls.json file. IMO, retargeting is meant to be completely transparent to the program relying on mbedls, which leaves control over whether or not to retarget in the hands of the user.

I can't see why anything relying on mbedls should disable retargeting, so to avoid confusion, I think it should be a separate function call.

bridadan: > IMO, retargeting is meant to be completely transparent to the program relying on mbedls, which leaves control over whether or not to retarget in the hands of the user.

@stevew817 You bring up a good point. Considering we make the same assumption about mocked targets, we should probably do the same with retargeting to keep the experience consistent. Any thoughts here @mazimkhan ?

mazimkhan: Hi @stevew817 I did not propose to pass a flag in the ```retarget()```. What I meant was to move the check ```if opts.skip_retarget:``` inside ```retarget()```. So that we don't have to check it every time we invoke ```retarget()```. But this will involve some more changes to make ```opts``` available inside ```retarget()```. May be have a accessor function to make it available globally. 
This is given that we have ```if opts.skip_retarget:```. We can't change a functionality and leave some options useless.
It would be worth to think if we require option ```opts.skip_retarget```. If we don't then we should also do the cleanup!

stevew817: @mazimkhan Making 'opts' globally available doesn't make a whole lot of sense, since mbedls isn't guaranteed to run commandline. Other mbed tools just call the create() method directly to get the mbeds object and run functions on that. No 'opts' are passed.

I'd be perfectly happy if we deprecate the skip_retarget option and just always apply retargeting. As said in my previous comment, if a user wants to stop retargeting, he/she should just remove/rename the retarget file or the relevant entry inside of it...

stevew817: @mazimkhan @bridadan Considering this PR has already taken more of my time then I expected it to: if this doesn't satisfy the mbed team, can I rely on you fixing the underlying issue one way or another? It really wasn't my duty to fix it to begin with...

Retargeting needs to work in all mbed tools. The way they currently interface with mbed-ls doesn't apply retargeting. Can you please fix this ASAP so us silicon partners can continue testing our future targets?

mazimkhan: @stevew817 Can you please revert your commit https://github.com/ARMmbed/mbed-ls/pull/146/commits/ccc88820ed9008685140ba82970a922c697921e5. Your previous changes are good to be merged.

stevew817: @mazimkhan Are you sure? That way, skip_retarget will stop working (as you correctly pointed out)

mazimkhan: It will only be applied when run from command line.

mazimkhan:
@stevew817 Please close this PR if satisfied with https://github.com/ARMmbed/mbed-ls/pull/150

stevew817:
Superseded by #150

PR #147, opened 2016-12-01
Add NUCLEO_F412ZG platform

bcostm:


bcostm:
I have one board on my desk and ID is correct too in the mbed.htm file when the board is plugged. FYI we are starting the tests on it...

PR #150, opened 2016-12-06
Fix retarget when mbedls is imported

mazimkhan:
Rework of https://github.com/ARMmbed/mbed-ls/pull/146

Retargeting requires explicit call to ```MbedLsToolsBase.retarget()``` after importing and calling ```create()```. This isn't convenient. Hence this change moves the call to ```MbedLsToolsBase.retarget()``` in ```MbedLsToolsBase.__init__()```. Since retargeting may be disabled by command line option ```--skip-retarger``` changes are made in constructors up to ```MbedLsToolsBase.__init__()``` to allow passing of arguments from ```main()``` to the base class.

Please review and test if it meets original PR requirements @bridadan @stevew817.

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -28,7 +28,7 @@ from lockfile import LockFailed, LockTimeout
 class MbedLsToolsBase:
     """ Base class for mbed-lstools, defines mbed-ls tools interface for mbed-enabled devices detection for various hosts
     """
-    def __init__(self):
+    def __init__(self, **kwargs):
         """ ctor
         """
         #extra flags
@@ -39,6 +39,9 @@ class MbedLsToolsBase:
         # Create in HOME directory place for mbed-ls to store information
         self.mbedls_home_dir_init()
         self.mbedls_get_mocks()
+        # skip retarget if specified to skip in arguments
+        if 'skip_retarget' not in kwargs or not kwargs['skip_retarget']:
+            self.retarget()
 
     # Which OSs are supported by this module
     # Note: more than one OS can be supported by mbed-lstools_* module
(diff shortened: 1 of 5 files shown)


mbed_lstools/lstools_base.py
```diff
@@ -28,7 +28,7 @@
 class MbedLsToolsBase:
     """ Base class for mbed-lstools, defines mbed-ls tools interface for mbed-enabled devices detection for various hosts
     """
-    def __init__(self):
+    def __init__(self, **kwargs):
```
bridadan: It looks like `skip_retarget` is the only keyword argument that's being used right now, so why not make it explicit here?

```python
def __init__(self, skip_retarget=False):
```

mbed_lstools/lstools_base.py
```diff
@@ -39,6 +39,9 @@ def __init__(self):
         # Create in HOME directory place for mbed-ls to store information
         self.mbedls_home_dir_init()
         self.mbedls_get_mocks()
+        # skip retarget if specified to skip in arguments
+        if 'skip_retarget' not in kwargs or not kwargs['skip_retarget']:
```
bridadan: Based on my above comment, this would just become the following:
```
if not skip_retarget:

stevew817:
LGTM on cursory read.

mbed_lstools/lstools_base.py
mazimkhan: If this extends in future we don't have to keep adding arguments in this class and the derived tree. It is fare to have a transparent mechanism between ```main()``` and this main class for the options in my opinion.

bridadan: That's fine, it's just harder to pick up on a first-time read of the code. Not that big of an issue though. And also the alternative is what the mbed OS tools do, which is not that great :stuck_out_tongue: 

Just look at this beatuy: https://github.com/ARMmbed/mbed-os/blob/master/tools/build_api.py#L365

PR #153, opened 2017-01-05
Add Realtek RTL8195AM

samchuarm:
- Add Realtek RTL8195AM


bridadan:
cc @mazimkhan @adbridge

Issue #119, opened 2016-08-09
Add support for public endpoint with boards, procuct codes etc.


PrzemekWirkus:
There is now a public https://www.mbed.com/api/v1/boards/ endpoint. It exposes all public boards. It contains information about product code, features, targets (classic, mbed os 5, yotta) and more. There is also similar endpoint on developer.mbed.org site but it requires token to get this information.

See `IOTWEB-1105` for more details.

PrzemekWirkus:
@mazimkhan @bridadan @adbridge @ConorPKeegan @screamerbg @sg- FYI

bridadan:
Ohhhhhh boy! This is gonna be sweet! :D

sg-:
Wowwwwwzzzzers 👍

jupe:
cool :+1: . is there also plain json API?

bridadan:
@jupe : https://www.mbed.com/api/v1/boards/?format=json

:)

bridadan:
FYI the api has moved to https://os.mbed.com/api/v3/platforms/

adbridge:
*Issue header has missing or incorrectly filled in fields! Please fix.*

mazimkhan:
Seems like this is implemented by #341. But I see there was a revert as well.

PR #157, opened 2017-02-08
[DO NOT MERGE] Refactor Proposal

bridadan:
It'll probably be easier to view the structure of this in my branch here: https://github.com/bridadan/mbed-ls/tree/refactor

This is a proposal to refactor mbedls. It has grown quickly and I think it's probably time to do a clean up!

There were a few motivations for doing this:
- File system accesses to mbed devices can be VERY SLOW on systems that have 10s of boards connected
    - This impacts CI automation significantly. We saw calls to mbedls that would take about 1 minute when we had 20 boards connected!
    - The current implementation makes no guarantees on when the file system will be accessed. It does it in multiple places when it isn't necessary to do so. This PR aims to consolidate all file system accesses in one place.
- Simplify the implementation and decouple the OS specific code from the shared logic
    - You'll notice many functions have a `# TODO add deprecation notice`. These functions can be trivially implemented in the client program, so it'd be good to remove this from the public API
    - This makes it simpler to maintain and easier to contribute
    - It also makes it easier to test the shared logic
- Prepare for the future as we add more mbeds to the ecosystem
    - There is now a dedicated "Platform Database" class that manages all platform entries
    - This could easily be expanded to use the online platform API (mentioned in #119)

### What's Done
- Transitioned logging to standard python [logging module](https://docs.python.org/2/library/logging.html)
- Consolidation of filesystem access in shared logic
- New abstract base class for OS specific implementation (MbedProvider)
- New class for managing platform names and target ID prefixes (MbedPlatformDatabase)
    - And tests!
- New class for managing mock file (MockManager)
    - And tests!
- New class for managing the discovery of all platform data (MbedLister)
    - And tests!
- Windows implementation (MbedProviderWindows)
- Linux implementation (MbedProviderLinux)
- Layer that is API-consistent with previous versions (MbedLsTool)

### What's NOT Done
- Break up the PR into actual commits to preserve history
- OS X implementation
- More complete debug output
- Being able to accurately set the logging level from the CLI
- Adding deprecated notices to API calls (pending review from @mazimkhan and @adbridge)
- Redo all code comments

Could you please take a look @mazimkhan and @adbridge 
FYI @c1728p9 

----

Below is a general description of how the mbedls flow works. You'll notice it's very similar to what was implemented previously, but now we can formally define it and enforce it! :smile: 

Call to list_mbeds (top level apis)
- arguments
    - **filesystem access** (choices below)
        - Before target ID check, most accurate but slowest option
            - Checks all boards are mounted before including them in results
        - After target ID check, provides more data but can lead to incorrect target IDs
            - Also ensures that the board is at least mounted
        - Disabled, very fast but provides only minimal information about the boards (and may be incorrect)
            - boards may not be mounted or ready for activity!
    - **filter targets by ID** (array of regular expressions)
        - Can be used to limit mbeds in results. Also can improve speed when used with **filesystem access** argument

1. Do initial fetch of target IDs/boards (should not access the mbed's filesystem)
	1. Relies on USBID string
1. If filesystem check is set to 'before target ID check'
	1. For each board, read filesystem and update all data (including target ID!)
1. If target ID filter is set
	1. remove non-matching mbeds from the list
1. If filesystem check is set to 'after target ID check'
    1. For each remaining board, read filesystem and update all data (including target ID!)
1. return list of mbed boards
1. Do all formatting in the common base class


jupe:
Cool! This sounds very good!
cc: @VeliMattiLahtela

bridadan:
@mazimkhan @adbridge bump! Any thoughts?

jupe:
any news about this ?

theotherjimmy:
Hey @jupe I'm going to take this PR from here as @bridadan no longer has time to support it. I expect to leave this branch as is, and open a PR that supersedes it.

theotherjimmy:
@jupe, looking on master now we have:

# From What's Done
- [x] Transitioned logging to standard python [logging module](https://docs.python.org/2/library/logging.html)
- [x] Consolidation of filesystem access in shared logic
- [ ] New abstract base class for OS specific implementation (MbedProvider) **NOT PLANNED**
- [x] New class for managing platform names and target ID prefixes (~Mbed~PlatformDatabase)
    - And tests!
- [x] New class for managing mock file (~MockManager~ PlatformDatabase)
    - And tests!
- [ ] New class for managing the discovery of all platform data (MbedLister) **NOT PLANNED**
    - And tests!
- [x] Windows implementation (MbedProviderWindows) (I never removed it)
- [x] Linux implementation (MbedProviderLinux) (Also never removed it)
- Layer that is API-consistent with previous versions (MbedLsTool)

# From What's NOT Done
- [x] Break up the PR into actual commits to preserve history
- [x] OS X implementation (I never removed it)
- [ ] More complete debug output (I'm not sure what this is about)
- [x] Being able to accurately set the logging level from the CLI (`-d` still works)
- [ ] Adding deprecated notices to API calls (pending review from @mazimkhan and @adbridge) **COMING SOON**
- [ ] Redo all code comments (I'm not sure what this is about either)

bridadan:
> More complete debug output (I'm not sure what this is about)
> Redo all code comments (I'm not sure what this is about either)

I had removed them when I started from scratch, this was just a note to myself to get them back in before merging anything. Not necessary in this case!

theotherjimmy:
As this is now all on master, I'm going to close this

bridadan:
Excellent work @theotherjimmy  :+1:

PR #158, opened 2017-02-10
Update lstools_base.py

sg-:
Board ID does not match the device database. Update to match DAPLink 0242 builds and newer


sg-:
Thanks @mazimkhan ! When will this be released in pip?

mazimkhan:
Welcome. Released version 1.2.12.

PR #159, opened 2017-02-15
Fix platform Id list

mazimkhan:
Added missing ```,``` after dict element.


bridadan:
Oh jeez, totally missed that missing comma sorry! We really need to get CI turned back on for this repo's PRs!

Issue #160, opened 2017-02-27
support for NRF52840_DK


nvlsianpu:
Nordic want to introduce nRF52840_DK port (http://www.nordicsemi.com/eng/Products/nRF52840-Preview-DK). mbed-os PR: https://github.com/ARMmbed/mbed-os/pull/3841. Interface firmware: (http://www.nordicsemi.com/eng/nordic/download_resource/60179/6/76665337). Reserved board ID <B>1102</B>.

nvlsianpu:
@pan- @screamerbg

theotherjimmy:
@nvlsianpu I'm closing this as #161 was merged.

PR #161, opened 2017-02-27
Add NRF52840 target to the list of platforms.

pan-:
This patch should solve #160 . 
@nvlsianpu could you verify this patch ?


nvlsianpu:
Verified successfully. This solved mentioned issue.

PR #166, opened 2017-03-28
[Silicon Labs] Update board IDs

stevew817:
Updated board IDs for Silicon Labs targets.


screamerbg:
We are aligned with Steven

stevew817:
@bridadan These targets aren't public yet, which is why we can get away with renaming them :)

stevew817:
@bridadan ARMmbed/mbed-os#4059 got merged, so this one should probably merge as well?

bridadan:
@stevew817 Fine with me, but I don't have merge powers :smile: That's up to @mazimkhan or @adbridge.

PR #169, opened 2017-03-31
Support NUMAKER_PFM_NANO130 and NUC240

cyliangtw:
Adding 2 new boards port for mbed OS. These 2 chips are based on Coretex-M0.
I also applied these 2 platforms into mbed database through Eric Yang.


cyliangtw:
@bridadan , Eric-Yang already inserted these 2 into mbed database.
Please check it again.

PR #175, opened 2017-05-02
Add support for LPC54114 & LPC54608

mmahadevan108:
Signed-off-by: Mahadevan Mahesh <Mahesh.Mahadevan@nxp.com>


mmahadevan108:
@maclobdell @0xc0170

bridadan:
It looks like the `LPC54114_M0` definition is missing in the mbed database and the entry for the `LPC54114_M4` doesn't seem to quite match that which is listed here. @maclobdell can you take a look?

mmahadevan108:
@bridadan @maclobdell I have changed LPC54114_M4 to LPC54114. I have also included support for LPC54608 in this PR.

maclobdell:
@mmahadevan108  Does the interface firmware enumerate two separate mass storage device drivers - one for each core?  We allocated 1054 for this platform.  1055 is available and we can allocate it too, but I want to confirm if to unique IDs are needed.   Thanks.

bridadan:
@maclobdell not sure what our convention is regarding how entries in the database match with the mbed-ls listing. The names in mbedls don't match any entries in the database at the moment, is this a requirement?

maclobdell:
@bridadan it should match.  Looks like https://github.com/ARMmbed/mbed-os/pull/4290 has been updated so it matches.   This platform has two targets, one for each core.

mmahadevan108:
It enumerates as 1. Initially I thought I would make it as 2 separate devices.

maclobdell:
the updates match the system.  I approve.  

Thanks.

PR #178, opened 2017-05-08
changed board name of ID:1022

ytsuboi:
CC: @toyowata


toyowata:
@ytsuboi Thanks for the PR. I have already changed the target name in platform page of the board ID 1022.

PR #179, opened 2017-05-09
DISCO_F413ZH: Add new target

arostm:
This is a new target to be added on mbed.


arostm:
Hi, when do you plan to merge it ?

PR #185, opened 2017-07-12
Fix error return codes on windows

c1728p9:
Only set ERRORLEVEL_FLAG if mount_point, target_id, serial_port or platform_name is missing from an mbed device's info. This allows devices without mbed.htm to be detected without errors, as long as the relevant information is available through another mechanism, such as details.txt.

This fixes mbed-ls returning an errorcode of -1 when a microbit is attached on a windows computer.


c1728p9:
@mazimkhan can you review this and merge it if it looks good?

Issue #167, opened 2017-03-29
python3 compatibility issues


jupe:
There is at least couple issues I've found so far which prevent to use mbedls from python3

**Environment:** win7
**python:** Python 3.6.1
**mbed-ls** tip of master (8363261250a5ff64af9b2574187d59d83a532cb8)

## Installation
```
Traceback (most recent call last):
  File "setup.py", line 38, in <module>
    long_description=read('README.md'),
  File "setup.py", line 33, in read
    return open(os.path.join(os.path.dirname(__file__), fname)).read()
  File "C:\Users\myuser\AppData\Local\Programs\Python\Python36\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 14188: character maps to <undefined>
```

## Usage
```
  File "C:\Users\myuser\AppData\Local\Programs\Python\Python36\lib\site-packages\mbed_ls-1.2.12-py3.6.egg\mbed_lstools\lstools_win7.py", line 213, in <listcomp>
    result += [d for d in self.get_dos_devices() if ven.upper() in d[1].upper()]
AttributeError: 'filter' object has no attribute 'upper'
```

jupe:
any news ?

jupe:
Installation seems to work better with Mac (+py3) but when I call mbed-ls it just stuck.

bridadan:
No news on this. If this is blocking you we need to get this work scheduled in.

jupe:
blocking: IOTSYST-2314 (not critical yet)

jupe:
created IOTMORF-981. any news?

mazimkhan:
@jupe If you are expecting any work on this. I would say I don't have any bandwidth for this. Please escalate and assign it in your team.

adbridge:
@jupe Likewise, unless it is scheduled into our sprints (which could be a while as it isn't a fundamental need for us) then I would suggest taking on the work yourselves, including rigorous testing to ensure backwards compatibility...

bridadan:
@jupe perhaps this should be taken into account with my proposed mbed-ls refactor here? https://github.com/ARMmbed/mbed-ls/pull/157

I have not tested with python 3 however this could be added to tests.

jupe:
Fix PR is available: https://github.com/ARMmbed/mbed-ls/pull/180

[...]

PR #192, opened 2017-08-21
Fix Python 3 compatibility issue

scartmell-arm:
- Changed the open() function to include an encoding argument for Python 3 support.

Fixes: python3 compatibility issues #167

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: python
 python:
   - "2.7"
-  - "3.4"
+  - "3.5"
+  - "3.6"
 
 install:
   - pip install coverage
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@
 
 # Description
 
-```mbed-ls``` is a _Python 2.7_ module that detects and lists mbed-enabled devices connected to the host computer. It will be delivered as a redistributable Python module (package) and command line tool.
+```mbed-ls``` is a Python (2 and 3) module that detects and lists mbed-enabled devices connected to the host computer. It will be delivered as a redistributable Python module (package) and command line tool.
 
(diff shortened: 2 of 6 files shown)


theotherjimmy:
@scartmell-arm I like it. Could you update the readme and the travis config to mention and test python 2 and 3, respectively?

bridadan:
@mazimkhan @adbridge  For some reason travis isn't running on PRs. Can you please enable this? Should be very simple to enable this from the Travis UI. Thanks!

bridadan:
So even though Travis isn't running at the moment, Python 3.4 is enabled in Travis now: https://github.com/ARMmbed/mbed-ls/blob/master/.travis.yml#L4

@theotherjimmy @scartmell-arm Perhaps we should add later versions of Python 3.x as well? Is that necessary?

theotherjimmy:
> perhaps we should add later versions of Python 3.x as well?

Agreed

> Is that necessary?

We should try to test with the most recent version in Ubuntu.

scartmell-arm:
Np.

> Is that necessary?

It's probably not necessary, but the test runs quick enough that there's no reason not to.

> We should try to test with the most recent version in Ubuntu.

The latest version of python3 is 3.6.x. The current and LTS releases of Ubuntu use 3.5.x and the next release in October will be 3.6.x.

mazimkhan:
Some how the repo was not active on travis. Activated it now. But it may only trigger with another push.

bridadan:
@mazimkhan Might also activate if you close and reopen the PR?

theotherjimmy:
Nailed it. We have all pythons passing. @mazimkhan Merge please?

Issue #56, opened 2016-03-01
Unmounted MBEDs are not listed


justbuchanan:
Right now on Linux, mbedls only shows MBEDs that are mounted to the filesystem.  It would be helpful if these were shown even if the mount point is blank.  This is especially useful for Linux distros that don't mount USB devices by default.

ciarmcom:
ARM Internal Ref: IOTSFW-2146

theotherjimmy:
I just merged #197 Clasing.

PR #197, opened 2017-09-19
List not mounted boards on Linux

theotherjimmy:
# Description

Adds the `-u`/`--list-unmounted` options to the command line of mbed-ls
that include boards that are not mounted in the mbed-ls output.

This is really only useful for systems that do not have a builtin
auto-mounter, such as a bare-bones linux install/distro.

Supersedes #56 

# TODO
 - [ ] Unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,8 @@ $ mbedls
 
 If you want to use ```mbedls``` in your toolchain, continuous integration or automation script and do not necessarily want to use the Python module ```mbed_lstools``` - this solution is for you.
 
+On some Linux systems, USB mass storage devices are not automatically mounted and do not show up when running `mbedls` by default. If you would like to include these not mounted devices in `mbedls` output, you can run mbed-ls with the `-u` option, such as `$ mbedls -u`.
+
 ## Exporting mbedls output to JSON
 
 You can export mbedls outputs to JSON format: just use the ```---json``` switch and dump your file on the screen or redirect to a file. It should help you further automate your processes.
--- a/mbed_lstools/lstools_linux_generic.py
+++ b/mbed_lstools/lstools_linux_generic.py
@@ -32,6 +32,7 @@ class MbedLsToolsLinuxGeneric(MbedLsToolsBase):
         # Since Ubuntu 15 DAplink serial port device can have pci- prefix, not only usb- one
         self.name_link_pattern = '((%s)-[0-9a-zA-Z_-]*_[0-9a-zA-Z]*-.*$)'% ('|'.join(["pci", "usb"]))
         self.mount_media_pattern = "^/[a-zA-Z0-9/]* on (/[a-zA-Z0-9/]*) "
+        self.list_unmounted = kwargs.get("list_unmounted", False)
 
         self.nlp = re.compile(self.name_link_pattern)
         self.hup = re.compile(self.hex_uuid_pattern)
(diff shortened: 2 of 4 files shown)


theotherjimmy:
@AnotherButler Could you review the docs changes? (last commit

@bridadan Could you review the code changes?

README.md
```diff
@@ -187,6 +187,8 @@ $ mbedls
 
 If you want to use ```mbedls``` in your toolchain, continuous integration or automation script and do not necessarily want to use the Python module ```mbed_lstools``` - this solution is for you.
 
+On some Linux systems, USB mass storage devices are not automatically mounted and do not show up when running `mbedls` by default. If you would like list these not mounted devices, you can run mbed-ls with the `-u` option: `$ mbedls -u`.
```
bridadan: I think you meant "If you would like to list" vs "If you would like list".

mbed_lstools/lstools_linux_generic.py
```diff
@@ -16,14 +16,16 @@
 """
 
 import re
+import subprocess
```
bridadan: Looks like `subprocess` got added here but isn't being used, can you remove it?

mbed_lstools/lstools_linux_generic.py
```diff
@@ -111,7 +114,12 @@ def list_mbeds(self):
                     mbed['target_id'] = mbed_htm_target_id
             mbed['target_id_usb_id'] = tid
             mbed['target_id_mbed_htm'] = mbed_htm_target_id
-            result.append(mbed)
+            if mbed['mount_point'] != None or self.list_unmounted:
+                result.append(mbed)
+            else:
+                logging.warn(
+                    "MBED with target id '%s' is connected, but not mounted. "
+                    "Use the '-u' flag to include it in the list." % mbed['target_id'])
```
bridadan: The whole code base currently isn't using the `logging` module. This should probably change in the future, but in the meantime I'd suggest using the `debug` function defined here: https://github.com/ARMmbed/mbed-ls/blob/408cd77254b6d476b730a18c1c142f5614b9be85/mbed_lstools/lstools_base.py#L545

You can see how it's used here: https://github.com/ARMmbed/mbed-ls/blob/92a85fd4aee17949da583fa61eff8c3bca95b9c0/mbed_lstools/lstools_linux_generic.py#L151

mbed_lstools/main.py
```diff
@@ -174,6 +181,9 @@ def get_mbedls_version():
     mbeds.debug(__name__, "mbed-ls ver. " + get_mbedls_version())
     mbeds.debug(__name__, "host: " +  str((mbed_lstools_os_info())))
 
+    if not opts.skip_retarget:
+        mbeds.retarget()
```
bridadan: I don't think this step belongs here, it's already handled in the lstools_base class: https://github.com/ARMmbed/mbed-ls/blob/a1aec616f93e0c2eb68889d38e8053918975b8bc/mbed_lstools/lstools_base.py#L77

README.md
theotherjimmy: You are correct.

mbed_lstools/lstools_linux_generic.py
theotherjimmy: Yes. Agreed.

mbed_lstools/lstools_linux_generic.py
theotherjimmy: 👍 I'll submit another PR for use of logging.

mbed_lstools/main.py
theotherjimmy: Heh, You're right.

theotherjimmy:
@bridadan I have pushed a working version which takes all of your review comments into account. Could you take a quick look?

theotherjimmy:
I left in the logging import. Just a sec.

mbed_lstools/lstools_linux_generic.py
```diff
@@ -23,7 +23,7 @@
 class MbedLsToolsLinuxGeneric(MbedLsToolsBase):
     """ MbedLsToolsLinuxGeneric supports mbed-enabled platforms detection across Linux family
     """
-    def __init__(self, **kwargs):
+    def __init__(self,  **kwargs):
```
theotherjimmy: How'd that sneak in?

Issue #163, opened 2017-03-06
Python 3 compatibility and reliability


hvenev:
```
Traceback (most recent call last):
  File "/usr/bin/mbedls", line 9, in <module>
    load_entry_point('mbed-ls==1.2.12', 'console_scripts', 'mbedls')()
  File "/usr/lib/python3.5/site-packages/mbed_lstools/main.py", line 218, in mbedls_main
    print(mbeds.get_string(border=not opts.simple, header=not opts.simple))
  File "/usr/lib/python3.5/site-packages/mbed_lstools/lstools_base.py", line 577, in get_string
    mbeds = self.list_mbeds_ext()
  File "/usr/lib/python3.5/site-packages/mbed_lstools/lstools_base.py", line 442, in list_mbeds_ext
    mbeds = self.list_mbeds()
  File "/usr/lib/python3.5/site-packages/mbed_lstools/lstools_linux_generic.py", line 84, in list_mbeds
    mbeds = self.get_detected(tids, disk_ids, serial_ids, mount_ids)
  File "/usr/lib/python3.5/site-packages/mbed_lstools/lstools_linux_generic.py", line 232, in get_detected
    mbed_mount_point = self.get_mount_point(mbed_dev_disk, mount_list)
  File "/usr/lib/python3.5/site-packages/mbed_lstools/lstools_linux_generic.py", line 299, in get_mount_point
    m = mmp.search(mount)
TypeError: cannot use a string pattern on a bytes-like object
```

The current approach to listing vfat filesystems is wrong:

```
$ mount /dev/sda42 /mnt/vfat
$ mount | grep vfat
/dev/sda42 on /mnt/vfat type ext4 (rw,relatime)
```

It is also unfixable:

```
$ mkdir '/tmp/a type proc'
$ mount -t tmpfs none '/tmp/a type proc'
$ mount
none on /tmp/a type proc type tmpfs (rw,relatime)
```

Some filesystems (e.g. btrfs, overlayfs) allow arbitrary strings in mount options. This can be used to create an ambiguity.

Use libmount instead (`/etc/mtab` escapes bad characters):

```
>>> import libmount
>>> ctx=libmount.Context()
>>> while True:
>>>     fs=ctx.mtab.next_fs()
>>>     if fs is None: break
>>>     print(repr((fs.target,fs.source)))
('/tmp/a type proc', 'none')
```

Note that this is still only marginally better. It still doesn't work on most random testcases (I've reported the bug to util-linux):

```
[...]

PR #202, opened 2017-09-22
Refactor mock and platform db into one class

### LINKED ISSUE #164: XDG Base Directory Specification compliance
The lock file should probably be placed under XDG_RUNTIME_DIR. Are there any other files created in the user's home directory?

Edit: you can use PyXDG

theotherjimmy:
Resolves #164  
Partial of #163 

This PR strips mocking functionality out of the MbedLsToolsBase class (what a name :( ) and makes a new module that is responsible for doing the database management. 

This new class, PlatformDatabase, accepts multiple files to build a database from and marks one of them as the primary_database. All new platforms are added to the primary db, and platform removals are done from all dbs. Only updates to the primary database are synced with the file system. When modifying a database, all accesses are synchronized with a file system lock (including all modifications of the mocks and the platform database in the user's home dir)

Coverage of this new file is 97%

### DIFF
--- a/mbed_lstools/platform_database.py
+++ b/mbed_lstools/platform_database.py
@@ -1,0 +1,341 @@
+"""
+mbed SDK
+Copyright (c) 2017 ARM Limited
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+"""Functions that manage a platform database"""
+
+import json
+import logging
+import re
+from collections import OrderedDict
+from copy import copy
+from io import open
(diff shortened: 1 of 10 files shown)


theotherjimmy:
I don't know why Circle is being so odd.

theotherjimmy:
@mazimkhan @bridadan Care to review

@bridadan This is a combination of your MockManager and PlatformDatabase classes from your refactor

mbed_lstools/platform_database.py
```diff
@@ -0,0 +1,341 @@
...
+            for entry in db.items():
+                yield entry
+
+    def all_ids(self):
+        return iter(self._keys)
+
+    def get(self, index, default=None):
+        """Standard lookup function. Works exactly like a dict"""
+        for db in self._dbs.values():
+            maybe_answer = db.get(index, None)
```
bridadan: No critique, just the variable name made me happy 😄

theotherjimmy: Yeah, it might not be an answer, so maybe!

theotherjimmy:
I'm going to track down this circle failure today. Then I'll update this PR if I can.

theotherjimmy:
@bridadan Could you check again to ensure that I did not break anything with my rampant rebasing?

theotherjimmy:
@bridadan I'm planning on it. After this refactor, there should be a few more. at that point, we should release version 2.0.

PR #203, opened 2017-09-23
Use colorlog when it's available

theotherjimmy:
Mbed LS will work without it, but your logs instantly become more readable when
you install colorlog

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,4 +52,9 @@ setup(name='mbed-ls',
       },
       install_requires=[
         "PrettyTable>=0.7.2",
-        "fasteners"])
+          "fasteners"
+      ],
+      extras_require = {
+          "colorized_logs": ["colorlog"]
+      }
+)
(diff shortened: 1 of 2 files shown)


theotherjimmy:
@bridadan @mazimkhan Would either of you like to take a look?

setup.py
```diff
@@ -52,4 +52,9 @@ def read(fname):
       },
       install_requires=[
         "PrettyTable>=0.7.2",
-        "fasteners"])
+          "fasteners"
+      ],
+      extras_require = {
```
bridadan: I'm not all that familiar with `extras_require`. Will it always try to install the packages and just not blow up if it fails?

theotherjimmy: It allows the `"mbed-ls[colorized_log]"`, but the code will Just Work (tm) if it can import the `colorlog` package.

bridadan: Cool 👍

PR #180, opened 2017-06-06
python3 fix

YifeiZuo01:
## Status
**IN DEVELOPMENT**

## Description
add decode to mount to support python3

## Relates to
* https://github.com/ARMmbed/mbed-flasher/issues/50
* #167

PR #205, opened 2017-09-25
Reapplying patch to fix python 3 support

bridadan:
This reapplies the change made in #180. It looks like the PR may have conflicted with #192, not entirely sure though.

@YifeiZuo01 Does this fix your issue? I don't have a Python 3 environment readily available to test.

### DIFF
--- a/mbed_lstools/lstools_linux_generic.py
+++ b/mbed_lstools/lstools_linux_generic.py
@@ -299,7 +299,7 @@ class MbedLsToolsLinuxGeneric(MbedLsToolsBase):
         mount_media_pattern = "^/[a-zA-Z0-9/]*/" + dev_name  + " on (/[a-zA-Z0-9_\-/]*) "
         mmp = re.compile(mount_media_pattern)
         for mount in mount_list:
-            m = mmp.search(mount)
+            m = mmp.search(mount.decode('utf-8'))
             if m and len(m.groups()):
                 return m.group(1)
         return None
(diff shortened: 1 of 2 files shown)


theotherjimmy:
@bridadan It looks like the tests pass this parameter incorrectly as a string, instead of bytes. Could you change that?

bridadan:
@theotherjimmy Updated. It looks a little funky but I think this may be the best way since it works in both Python 2 and 3.

YifeiZuo01:
Still, connected with devices:
    `m = mmp.search(mount.encode('utf-8').decode('utf-8'))`
`AttributeError: 'bytes' object has no attribute 'encode' `

theotherjimmy:
@bridadan This patch does not work, and the prior version:
```diff 
diff --git a/mbed_lstools/lstools_linux_generic.py b/mbed_lstools/lstools_linux_generic.py
index d0f15fb..a0569af 100644
--- a/mbed_lstools/lstools_linux_generic.py
+++ b/mbed_lstools/lstools_linux_generic.py
@@ -299,7 +299,7 @@ class MbedLsToolsLinuxGeneric(MbedLsToolsBase):
         mount_media_pattern = "^/[a-zA-Z0-9/]*/" + dev_name  + " on (/[a-zA-Z0-9_\-/]*) "
         mmp = re.compile(mount_media_pattern)
         for mount in mount_list:
-            m = mmp.search(mount)
+            m = mmp.search(mount.decode('utf-8'))
             if m and len(m.groups()):
                 return m.group(1)
         return None
```
Works correctly. I think that's why I told you to change the **test**.

bridadan:
Thanks for testing y'all, sorry it was late in the day and I got lazy.

I'll patch this up today!

theotherjimmy:
Hmmmmmm.... Just a moment.

theotherjimmy:
Wait for testing again...

theotherjimmy:
👍

PR #210, opened 2017-09-26
Fixing how plat_db is used in windows

bridadan:
Hooray for not passing in `self` variables to member functions!


theotherjimmy:
BAM!

theotherjimmy:
🍾

PR #211, opened 2017-09-27
Refactor fs and plat_db access into base class

theotherjimmy:


theotherjimmy:
@bridadan @mazimkhan Could you review?

mbed_lstools/lstools_base.py
```diff
@@ -48,7 +51,7 @@ class MbedLsToolsBase:
     MBED_HTM_NAME = 'mbed.htm'
     ERRORLEVEL_FLAG = 0
 
-    def __init__(self, **kwargs):
+    def __init__(self, list_unmounted=False, **kwargs):
```
bridadan: @theotherjimmy this should probably be in `list_mbeds`, not in the constructor, wouldn't you say? Seems a bit strange to have to reconstruct the object to change this option, especially since it's only used by the `list_mbeds` function.

mbed_lstools/lstools_base.py
```diff
@@ -61,10 +64,52 @@ def __init__(self, **kwargs):
...
+                     not self.list_unmounted):
+                    logger.warning(
+                        "MBED with target id '%s' is connected, but not mounted. "
+                        "Use the '-u' flag to include it in the list.",
+                        device['target_id'])
+            else:
+                htm_target_id = self.read_htm_target_id(device['mount_point'])
+                if htm_target_id:
+                    logging.debug("Found more accurate target id %s -> %s",
+                                  device['target_id'], htm_target_id)
```
bridadan: I _think_ this log may be a bit misleading, as the target ID returned from the USB info can be the same as the target ID returned from the htm file.

I'd recommend only printing out this log only if they differ.

mbed_lstools/lstools_darwin.py
```diff
@@ -26,134 +26,86 @@
...
-        # put together all of that info into the expected format:
-        result =  [
+        return [
             {
                 'mount_point': mounts[v],
                 'serial_port': volumes[v]['tty'],
-                  'target_id': self.target_id(volumes[v]),
-              'platform_name': self.platform_name(self.target_id(volumes[v]))
-            } for v in valid_volumes
+                  'target_id': volumes[v].get('serial')
```
bridadan: Looks like indentation got a little messed up here

mbed_lstools/lstools_darwin.py
```diff
@@ -26,134 +26,86 @@
...
+            } for v in set(volumes.keys()) & set(mounts.keys())
```
bridadan: Nit: I think its more Pythonic/consistent with the other tools to use `and` vs `&`. Not a big deal, and I could be wrong.

mbed_lstools/lstools_linux_generic.py
```diff
@@ -16,290 +16,85 @@
...
-        return self.get_dev_by_id_process(lines, retval)
-
-    def get_mounts(self):
+        dir = join("/dev", device_type, "by-id")
+        if isdir(dir):
+            to_ret = dict(self._hex_ids([join(dir, f) for f in listdir(dir)]))
+            logger.debug("Found %s devices by id %r", device_type, to_ret)
+            return to_ret
+        else:
+            logger.error("Could not get %s devices by id", device_type)
```
bridadan: Is the failure mode here that the `/dev/{type}` directory doesn't exist? Does this mean their distribution is handling disk/serial devices in nonstandard way? Should they raise an issue on mbed-ls in that case so we could potentially add support?

mbed_lstools/lstools_linux_generic.py
```diff
@@ -16,290 +16,85 @@
 """
 
 import re
```
bridadan: This file is a thing of beauty, well done! 👍

mbed_lstools/lstools_darwin.py
```diff
@@ -26,134 +26,86 @@
...
-                mount_point = disk['MountPoint']
-            r[disk['DeviceIdentifier']] = mount_point
-
-        return r
-
+        return {disk['DeviceIdentifier']: disk.get('MountPoint', None)
+                for disk in disks['AllDisksAndPartitions']}
 
-    def get_mbed_volumes(self):
+    def get_volumes(self):
```
bridadan: I noticed in the linux class, you moved all functions that weren't `find_candidates` to have the private `_` prefix. Do you want to something similar here?

mbed_lstools/lstools_win7.py
```diff
@@ -28,86 +28,52 @@
...
-
-        mbeds = []
-        for mbed in self.discover_connected_mbeds():
-            d = {}
-            d['mount_point'] = mbed[0] if mbed[0] else None
-            d['target_id'] = mbed[1] if mbed[1] else None
-            d['serial_port'] = mbed[2] if mbed[2] else None
-            d['platform_name'] = mbed[3] if mbed[3] else None
-            d['target_id_usb_id'] = mbed[4] if mbed[4] else None
-            d['target_id_mbed_htm'] = mbed[5] if mbed[5] else None
```
bridadan: Glad to see this mess go 😛

mbed_lstools/lstools_win7.py
```diff
@@ -175,38 +134,25 @@ def get_mbeds(self):
...
-        @param mnt mbed mount point (disk / drive letter)
-        @param target_usb_id mbed target USB ID
-        @return Function returns (<target_id>, <htm_target_id>)
-        @details Helper function
-        """
-        mbed_htm_target_id = self.get_mbed_htm_target_id(mnt)
-        # Deducing mbed-enabled TargetID based on available targetID definition DB.
-        # If TargetID from USBID is not recognized we will try to check URL in mbed.htm
-        mbed_id = mbed_htm_target_id if mbed_htm_target_id is not None else target_usb_id
-        return mbed_id, mbed_htm_target_id
```
bridadan: Awesome, get this outta here 👍

test/os_linux_generic.py
```diff
@@ -29,433 +29,321 @@ class LinuxPortTestCase(unittest.TestCase):
...
-            "/media/usb0",
-            "/dev/ttyACM1",
-            "usb-MBED_microcontroller_0240020152986E5EAF6693E6-0:0 -> ../../sdb"
-          ],
+        mbed_det = self.find_candidates_with_patch(
+            self.mount_list_1, self.link_dict_1, self.listdir_dict_1)
+        self.assertIn({
+            'mount_point': "/media/usb0",
+            'serial_port': "/dev/ttyACM1",
+            'target_id': "0240020152986E5EAF6693E6"
```
bridadan: Nit: standardizing on `"` or `'` would be nice for consistency, not a hard requirement for this PR though.

mbed_lstools/lstools_base.py
theotherjimmy: I'm only changing internal APIs in this PR. This is an external API.

mbed_lstools/lstools_base.py
theotherjimmy: This is a debug log. Users should not see this unless they ask for debug output.

mbed_lstools/lstools_darwin.py
theotherjimmy: I'll change the others, they were wrong before.

mbed_lstools/lstools_darwin.py
theotherjimmy: I'll change it.

mbed_lstools/lstools_darwin.py
theotherjimmy: Yeah. I also dropped the superfluous `get` prefixes. When everything gets, nothing gets.

mbed_lstools/lstools_linux_generic.py
theotherjimmy: Thank you. 😊

mbed_lstools/lstools_linux_generic.py
theotherjimmy: > Is the failure mode here that the /dev/{type} directory doesn't exist? 

Yes

>does this mean their distribution is handling disk/serial devices in nonstandard way?

Yes

> Should they raise an issue on mbed-ls in that case so we could potentially add support?

Yes and no. They could encourage their distribution to use sane udev rules or switch to a distribution that does. They _could_ open an issue as well, but I can't think of a distribution that does not do this series of udev links.

mbed_lstools/lstools_win7.py
theotherjimmy: It was a mess indeed.

mbed_lstools/lstools_win7.py
theotherjimmy: 😊

mbed_lstools/lstools_base.py
theotherjimmy: The fact also remains that we did find a more accurate id, which happens to be the same id.

theotherjimmy: For debugging purposes, this log output *has been helpful*. Just seeing that it did this step is good.

mbed_lstools/lstools_linux_generic.py
theotherjimmy: @bridadan Do you want the error message changed?

theotherjimmy:
@bridadan I ready for review again.

mbed_lstools/lstools_base.py
bridadan: According to the diff you've added an optional parameter here. So adding it to `list_mbeds` wouldn't be a disruptive change. It'd be a "minor" bump vs a "major"

mbed_lstools/lstools_base.py
```diff
@@ -61,10 +64,52 @@ def __init__(self, **kwargs):
...
+                if  (device['target_id'] and device['serial_port'] and
+                     not self.list_unmounted):
+                    logger.warning(
+                        "MBED with target id '%s' is connected, but not mounted. "
+                        "Use the '-u' flag to include it in the list.",
+                        device['target_id'])
+            else:
+                htm_target_id = self.read_htm_target_id(device['mount_point'])
+                if htm_target_id:
+                    logging.debug("Found htm target id, %s, for usb target id %s",
```
bridadan: I like this better, nice.

One question, the previous version also included the target_id based on the usb: https://github.com/ARMmbed/mbed-ls/blob/master/mbed_lstools/lstools_win7.py#L58

It'd probably be good to not drop this, are you ok adding this back in? Maybe the `find_candidates` calls should just assign to `target_id_usb_id` and this function here can assign the `target_id`? Since that's what functionally happens.

mbed_lstools/lstools_base.py
theotherjimmy: Actually, this parameter is already accepted. I just made it explicit. It was parsed from `kwargs` before.

mbed_lstools/lstools_base.py
bridadan: Oh god, this is such small nit I'm sorry:

"Found htm target id, %s, for usb target id %s"

I think the `,` between `target id` and `%s` is unneeded.

But feel free to completely ignore this.

theotherjimmy: Sure, that's a good idea.

theotherjimmy: Do I have to call it `target_id_usb_id`? it's redundant to have 2 `id`s in one string. What do you think of `usb_target_id`? or `usb_id`? (everything's a target 😉 )

mbed_lstools/lstools_base.py
bridadan: Ah I see. Ok, it makes more sense to me for it being in `list_mbeds`, but we could always revisit that. Don't need to hold up this PR and all of its other goodness.

mbed_lstools/lstools_base.py
bridadan: For legacy sake, yes unfortunately. I agree it's redundant.

mbed_lstools/lstools_linux_generic.py
```diff
@@ -16,290 +16,89 @@
...
-    def get_mounts(self):
+        dir = join("/dev", device_type, "by-id")
+        if isdir(dir):
+            to_ret = dict(self._hex_ids([join(dir, f) for f in listdir(dir)]))
+            logger.debug("Found %s devices by id %r", device_type, to_ret)
+            return to_ret
+        else:
+            logger.error("Could not get %s devices by id. "
+                         "This could be because your Linux distribution "
+                         "dose not use udev, or does not create /dev/<type>/by-id "
```
bridadan: `dose` -> `dose` plz 😄 

Also, can you substitute the actual `device_type` variable for the `"<type>"` in this line? Just so its more clear for users.

mbed_lstools/lstools_linux_generic.py
```diff
@@ -16,290 +16,89 @@
...
+                         "symlinks. Please submit an issue to github.com/"
+                         "armmmeb/mbed-ls.", device_type)
```
bridadan: `armmmeb/mbed-ls` -> `armmbed/mbed-ls`

theotherjimmy:
@bridadan Another Another!

theotherjimmy:
Oh wow. that testing was fast.

PR #212, opened 2017-09-28
Remove duplicate log when color logging

theotherjimmy:


theotherjimmy:
@bridadan I figured it out!

PR #213, opened 2017-09-28
Deprecate fancy printing and internal methods

theotherjimmy:


mbed_lstools/lstools_base.py
```diff
@@ -308,23 +324,27 @@ def get_string(self, border=False, header=True, padding_width=1, sortby='platfor
 
     # Private functions supporting API
 
+    @deprecated("Using internal functions is deprecated.")
```
bridadan: For this particular deprecated message I think I'd give a bit more of a warning/reason.

"This function will be removed from the public API in future releases. Please use 'list_mbeds' instead."

test/listing.py
```diff
@@ -66,20 +66,6 @@ def test_list_mbeds_ext_mandatory_fields_exist(self):
...
-        for p in platforms:
-            self.assertIs(type(p), str)
-
-    def test_list_platforms_ext(self):
-        platforms = self.mbeds.list_platforms_ext()
-        self.assertIs(type(platforms), dict)
-        for p in platforms:
-            self.assertIs(type(p), str)
-            self.assertIs(type(platforms[p]), int)
-            self.assertTrue(platforms[p] >= 0)
```
bridadan: Even if functions are deprecated, should we still test them? Seems like when they're removed it'd be appropriate to stop testing them.

If you feel strongly the other way then no worries, just my initial reaction.

test/mbedls_toolsbase.py
```diff
@@ -63,10 +63,6 @@ def test_list_mbeds(self):
         self.assertEqual(to_check[0]['target_id'], "0241BEEFDEAD")
         self.assertEqual(to_check[0]['platform_name'], 'foo_target')
 
-    def test_list_manufacture_ids(self):
-        table_str = self.base.list_manufacture_ids()
-        self.assertTrue(isinstance(table_str, basestring))
```
bridadan: Ditto here about testing

mbed_lstools/lstools_base.py
```diff
@@ -128,6 +139,7 @@ def mock_manufacture_id(self, mid, platform_name, oper='+'):
         else:
             raise ValueError("oper can only be [+-]")
 
+    @deprecated("Fancy target listing is out of scope of mbedls.")
```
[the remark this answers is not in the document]
theotherjimmy: That is snarky, but I would not say that it's less technical. I'm going to change it.

test/listing.py
theotherjimmy: I guess. I don't feel strongly either way.

theotherjimmy:
@bridadan Ready for review.

PR #220, opened 2017-10-02
Fix win case com port

bridadan:
It was something quite silly!

Seems funny that it didn't cause an exception...


theotherjimmy:
Right?

PR #223, opened 2017-10-04
Add NUCLEO_L433RC_P and NUCLEO_F722ZE boards

bcostm:
Add new NUCLEO boards.

Issue #222, opened 2017-10-03
Missing platforms in platform_database.py


maclobdell:
When comparing the list of supported platforms in mbed ls release 1.2.14 verses now on the master branch, the following platforms are missing.  Perhaps the new platform_database.py file was populated with an old list?

All these are missing.  Please add back!

"0743": "DISCO_F413ZH",
"0764": "DISCO_L475VG_IOT01A",
"0823": "NUCLEO_L496ZG",
"0833": "DISCO_L072CZ_LRWAN1",
"1114": "LPC1114",
"1237": "UBLOX_EVK_NINA_B1",
"1305": "NUMAKER_PFM_M2351",
"1306": "NUMAKER_PFM_NANO130",
"1307": "NUMAKER_PFM_NUC240",
"2040": "EFM32PG12_STK3402",
"2041": "TB_SENSE_12",
"2045": "TB_SENSE_1",
"5003": "ARM_MPS2_M0P",
"5501": "GR_LYCHEE",
"7011": "TMPM066",
"7402": "MBED_BR_HAT",
"C005": "MTM_MTCONNECT04S",
"C030": "UBLOX_C030_U201",
"C031": "UBLOX_C030_N211",
"C032": "UBLOX_C030_R404M",
"C033": "UBLOX_C030_R410M",
"C034": "UBLOX_C030_S200",
"C035": "UBLOX_C030_R3121",

bridadan:
@maclobdell I just double checked my PR #224, all the above (and then some!) have been added back.

PR #224, opened 2017-10-04
Restoring all platforms to platform database

bridadan:
This should solve #222 and replace #223.


bridadan:
Oops @theotherjimmy just saw you self-assigned #222, hopefully you don't mind 😛

theotherjimmy:
Heh, I don't mind when you do work for me.

bridadan:
I figured 😄

PR #233, opened 2017-10-11
Opt into debug logging

theotherjimmy:
instead of opt out


theotherjimmy:
sorry @c1728p9 This will fix it for you.

PR #237, opened 2017-10-12
Removing incorrect warning about unmounted devices on windows

bridadan:
Noticed this extra (and incorrect) logging in the logs posted here: https://github.com/ARMmbed/greentea/issues/241#issuecomment-336115932

The particular warning really only applies to Mac and Linux since they have a similar concept of mounted/unmounted. It does not apply to Windows.

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -57,10 +57,6 @@ class MbedLsToolsBase(object):
 
     __metaclass__ = ABCMeta
 
-    # Which OSs are supported by this module
-    # Note: more than one OS can be supported by mbed-lstools_* module
-    os_supported = []
-
     # Directory where we will store global (OS user specific mocking)
     HOME_DIR = expanduser("~")
     MOCK_FILE_NAME = '.mbedls-mock'
@@ -134,7 +130,8 @@ class MbedLsToolsBase(object):
             if  ((not device['mount_point'] or
                   not self.mount_point_ready(device['mount_point'])) and
                  not self.list_unmounted):
-                if  (device['target_id_usb_id'] and device['serial_port']):
+                if  (device['target_id_usb_id'] and device['serial_port']
+                    and self.__class__.__name__ != 'MbedLsToolsWin7'):
                     logger.warning(
                         "MBED with target id '%s' is connected, but not mounted. "
                         "Use the '-u' flag to include it in the list.",
--- a/test/os_win7.py
+++ b/test/os_win7.py
@@ -50,8 +50,12 @@ class Win7TestCase(unittest.TestCase):
         _winreg.SetValueEx.reset_mock()
         _winreg.SaveKey.reset_mock()
 
-    def test_os_supported(self):
-        pass
+    '''
+    This test case is necessary as some logging output is irrelavent when the
+    tool is running on Windows and the class name is used to detect this
+    '''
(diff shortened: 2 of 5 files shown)


theotherjimmy:
@bridadan I'm not convinced that this is correct, and I don't like the implementation. That is: I'm not convinced that windows will never end up in a state where a board is only connected via CDC and the MSD part will not have a drive letter. 

Could we find a way to check this that's not "look at the class"? That feels like a hack to me.

test/os_win7.py
```diff
@@ -50,8 +50,12 @@ def setUp(self):
...
         _winreg.SaveKey.reset_mock()
 
-    def test_os_supported(self):
-        pass
+    '''
+    This test case is necessary as some logging output is irrelavent when the
+    tool is running on Windows and the class name is used to detect this
+    '''
+    def test_class_name(self):
+        self.assertEqual(self.lstool.__class__.__name__, 'MbedLsToolsWin7')
```
theotherjimmy: This test does not cover the behavior that is says it does in the string/comment above.

mbed_lstools/lstools_base.py
```diff
@@ -134,7 +130,8 @@ def list_mbeds(
             if  ((not device['mount_point'] or
                   not self.mount_point_ready(device['mount_point'])) and
                  not self.list_unmounted):
-                if  (device['target_id_usb_id'] and device['serial_port']):
+                if  (device['target_id_usb_id'] and device['serial_port']
+                    and self.__class__.__name__ != 'MbedLsToolsWin7'):
```
theotherjimmy: Is there a way to do this without using dunderscore (`__`) variables?

theotherjimmy:
@jeromecoutant I'm not trying to say that this output is correct. What I want to convey is that there might be a case where this output would be correct on windows, and we should endeavor to find a method of filtering these messages that still warns about poorly behaved boards.

jeromecoutant:
About the warning I have, I think they are due to my mounted network drive.

mbed_lstools/lstools_base.py
bridadan: Yeah I can do it another way, I'll revise in a bit.

bridadan:
@theotherjimmy 

> I'm not convinced that windows will never end up in a state where a board is only connected via CDC and the MSD part will not have a drive letter

I can't say with 100% certainty that this case doesn't happen on Windows. **However** I am 100% certain that there is a case where I don't have any boards connected and I get warnings. The warning is the following:

```
WARNING:mbedls.lstools_base:MBED with target id '0240000032044e4500257009997b00386781000097969900' is connected, but not mounted. Use the '-u' flag to include it in the list.
```

This warning is incorrect and confusing because I in fact don't have any boards connected.

And what's even more exciting is if I run mbedls with `-u`, it shows that a board is connected when really it isn't. If I try and open the listed COM port it throws an error (because it doesn't exist).

So I think we need to revisit the behavior on Windows.

theotherjimmy:
After an offline discussion with @bridadan We came to the conclusion that windows implementation is returning things that are not candidates. In particular, windows seems to keep around any board that was ever connected in the registry. We think that filtering on boards that are currently connected will fix this problem at the source.

bridadan:
Gonna open up a new PR with a more accurate description of the problem

PR #240, opened 2017-10-12
Run generic setup before anything else in main

theotherjimmy:


theotherjimmy:
@bridadan No more accidentally the root logger!

PR #242, opened 2017-10-12
Allowing testing on Windows

bridadan:
Previously the Linux implementation selectively imported `readlink` from the `os` module. Now we just import `os` and use `os.readlink` within the Linux implementation when we're ready. But this prevents the error at input time.

### DIFF
--- a/test/os_linux_generic.py
+++ b/test/os_linux_generic.py
@@ -80,8 +80,8 @@ class LinuxPortTestCase(unittest.TestCase):
 
     def find_candidates_with_patch(self, mount_list, link_dict, listdir_dict):
         with patch('mbed_lstools.linux.MbedLsToolsLinuxGeneric._run_cli_process') as _cliproc,\
-             patch('mbed_lstools.linux.readlink') as _readlink,\
-             patch('mbed_lstools.linux.listdir') as _listdir,\
+             patch('os.readlink') as _readlink,\
+             patch('os.listdir') as _listdir,\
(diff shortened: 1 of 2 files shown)


test/os_linux_generic.py
```diff
@@ -80,7 +80,7 @@ def test_get_mount_point_ext(self):
 
     def find_candidates_with_patch(self, mount_list, link_dict, listdir_dict):
         with patch('mbed_lstools.linux.MbedLsToolsLinuxGeneric._run_cli_process') as _cliproc,\
-             patch('mbed_lstools.linux.readlink') as _readlink,\
+             patch('mbed_lstools.linux.os.readlink') as _readlink,\
```
theotherjimmy: Could this also be `os.readlink`? would that be more readable?

theotherjimmy: Instead of `mbed_lstools.linux.os.readlink`

bridadan: Yeah sure, will do

theotherjimmy:
Man that bot does not like you're code! 

I think it's being irritated by `listdir` being used directly, instead of through `os.listdir`.

theotherjimmy:
Don't worry, I like your code, even if the bot does not.

bridadan:
Lol just had to fix it, no biggie. T'was wrong code anyway

theotherjimmy:
You can now rebase/merge your other PR!

PR #249, opened 2017-10-17
Get tests passing on Windows (and other cleanup)

bridadan:
This PR does a number of things regarding testing:

- Needed to patch `abspath` behavior so Linux tests would pass on Windows
- Change how temporary files are created since they are treated differently on Windows. See the following for more information (https://docs.python.org/2/library/tempfile.html#tempfile.NamedTemporaryFile)
- Remove old, incorrect, and/or pointless tests


theotherjimmy:
Thanks! As for the coverage going down, it looks like it was all false coverage anyway. So this 73% is more accurate than the prior 76%

PR #258, opened 2017-11-01
Add TMPM46B

soramame21:
toshiba board TMPM46B


theotherjimmy:
Thanks for the check @bridadan.

Issue #264, opened 2017-11-09
Targets not detected on Windows 7


MarceloSalazar:
Latest version of mbedls (1.3.3) doesn't detect targets (K64, ST Disco) on Windows 7.

Downgrading to 1.3.1 works ok.

@theotherjimmy

MarceloSalazar:
I guess is related to this https://github.com/ARMmbed/mbed-ls/issues/263

bridadan:
@MarceloSalazar Yes most likely, since the change happened in 1.3.2. I'll be putting up a patch to fix this momentarily, thanks for your patience.

theotherjimmy:
Thanks Brian. This has been released for a while.

PR #265, opened 2017-11-10
Remove use of 'fsutil' during Windows detection

### LINKED ISSUE #263: mbedls v1.3.3 requires admin rights on Window 7
Due to calling ```fsutil.exe```.
Can someone in team add this to the document?

bridadan:
Should fix #263 and #264.

In #241, I modified the detection process for Windows to use the [fsutil](https://technet.microsoft.com/en-us/library/cc753059(v=ws.11).aspx) program that ships with Windows. It provides a very fast way to detect what drive letters are currently available on the system. However, I missed this very important fact (spelled out in plain ol' English on their doc page):

> You must be logged on as an administrator or a member of the Administrators group to use fsutil. The fsutil command is quite powerful and should be used only by advanced users who have a thorough knowledge of Windows operating systems.

😞 

This PR changes the drive detection to the following:
- Take the results that are returned from the registry query
    - This includes mountpoints that may or may not be present on the system
- Loop over each mountpoint and call the Windows `dir` command
    - If it fails, don't include it in the returned list
    - The downside to this approach is we have to make more system calls
         - "But why not just use `os.path.exists` ?!?!"
         - If you call that function on a drive that is in the ejected state (happens directly after flashing), Python will throw a **blocking error box** and halt, and the workaround [isn't that great](https://stackoverflow.com/a/4790310/1493711)
    - The upside is this doesn't require administrator privileges

FYI @toyowata @MarceloSalazar

### DIFF
--- a/test/os_win7.py
+++ b/test/os_win7.py
@@ -19,6 +19,7 @@ limitations under the License.
 
 import unittest
 import sys
+import os
 from mock import MagicMock, patch
 
 # Mock the winreg and _winreg module for non-windows python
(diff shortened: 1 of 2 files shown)


bridadan:
Could you please review @theotherjimmy ?

bridadan:
Well that's weird, works just fine on my Windows box. I'll try it in my VM.

test/os_win7.py
```diff
@@ -145,18 +146,40 @@ def test_one_nucleo_dev(self):
...
                 'target_id_usb_id': u'000440035522'
             }
-            _cliproc.return_value = (b'\nDrives: C:\ D:\ F:\ Z:\ \n', None, 0)
-            devices = self.lstool.find_candidates()
-            self.assertIn(expected_info, devices)
-            self.assertNoRegMut()
 
-            _cliproc.return_value = (b'\nDrives: C:\ F:\ Z:\ \n', None, 0)
+
+            def _mounted_drives_check(drives):
```
theotherjimmy: Could you move this function outside of the test?

theotherjimmy:
Thanks for the update. @MarceloSalazar You could use a new release with this change set correct?

bridadan:
> Thanks for the update. @MarceloSalazar You could use a new release with this change set correct?
@theotherjimmy my understanding is he has a workshop to put on so yes to a new release!

MarceloSalazar:
Thanks @bridadan @theotherjimmy for the quick fix!
I confirm v1.3.4 works well on Windows 7 :)

toyowata:
@bridadan @theamirocohen me too!! Thanks for your fix!  :-)

bridadan:
Great to hear, thanks for testing @MarceloSalazar and @toyowata!

PR #269, opened 2017-11-20
Fix retargeting bug when using mbedls.json

hug-dev:
A NamesError exception is caused because some variables are not defined.
Changes the debug output. Fixes #268.

Issue #268, opened 2017-11-17
Retargeting with mbedls.json causes a NameError exception


hug-dev:
On ARM_CM3DS_MPS2 platform, with no `mbedls.json` file the output is:
```
$ mbedls
+---------------+----------------------+--------------------------------+-------------+--------------------------+-----------------+
| platform_name | platform_name_unique | mount_point                    | serial_port | target_id                | daplink_version |
+---------------+----------------------+--------------------------------+-------------+--------------------------+-----------------+
| unknown       | None[0]              | /media/              /V2M-MPS2 | unknown     | 50040200074D652F3828F333 | 0001            |
+---------------+----------------------+--------------------------------+-------------+--------------------------+-----------------+
```
The `mbedls.json` file is put in the current folder:
```
{
    "50040200074D652F3828F333": {
        "serial_port": "/dev/ttyUSB0"
    }
}
```
`mbedls` command now produces the `NameError` exception:
```
$ mbedls
Traceback (most recent call last):
  File "/home/              /dev/tmp_cm3ds/patch_mbedls/venv/bin/mbedls", line 11, in <module>
    load_entry_point('mbed-ls==1.3.4', 'console_scripts', 'mbedls')()
  File "/home/              /dev/tmp_cm3ds/patch_mbedls/venv/local/lib/python2.7/site-packages/mbed_ls-1.3.4-py2.7.egg/mbed_lstools/main.py", line 261, in mbedls_main
    ret_code = args.command(mbeds, args)
  File "/home/              /dev/tmp_cm3ds/patch_mbedls/venv/local/lib/python2.7/site-packages/mbed_ls-1.3.4-py2.7.egg/mbed_lstools/main.py", line 110, in print_table
    return print_mbeds(mbeds, args, False)
  File "/home/              /dev/tmp_cm3ds/patch_mbedls/venv/local/lib/python2.7/site-packages/mbed_ls-1.3.4-py2.7.egg/mbed_lstools/main.py", line 97, in print_mbeds
    devices = mbeds.list_mbeds(unique_names=True, read_details_txt=True)
  File "/home/              /dev/tmp_cm3ds/patch_mbedls/venv/local/lib/python2.7/site-packages/mbed_ls-1.3.4-py2.7.egg/mbed_lstools/lstools_base.py", line 162, in list_mbeds
    device['target_id'], mbeds[i])
NameError: global name 'mbeds' is not defined
```

The reason of this has been identified and a pull request is soon going to be made to fix this.

PR #271, opened 2017-11-21
Add test for missing retarget coverage

bridadan:
This PR adds a test for the issue that was raised in #268.

This PR will fail as is. When #269 is merged, I will rebase it so we can make sure it passes.


theotherjimmy:
It's great to know that this fails without #269!

bridadan:
Rebased!

Issue #277, opened 2017-12-08
`get_mbed_com_port` API is missing


jupe:
`get_mbed_com_port` is dropped in some point but mbed-flasher rely on it still. 
This need to bring back including regression test for it. This is also mentioned in README.md still.

```
Instance of 'MbedLsToolsWin7' has no 'get_mbed_com_port' member
Instance of 'MbedLsToolsLinuxGeneric' has no 'get_mbed_com_port' member
Instance of 'MbedLsToolsDarwin' has no 'get_mbed_com_port' member
```

bridadan:
@jupe That was only ever and API in the Windows implementation of the mbed-ls backend. It really was never meant to be consumed, it was an internal API. You should be able to get that same information from the public API, is it possible to switch mbed-flasher to the supported API?

jupe:
What is supported API's ? Would be good to have documented them first :)

jupe:
but sure we could change mbed-flasher side as well. I hope that documentation and regression tests (especially API tests) would be in place sooner or later so that we doesn't ended up in this kind of case in future anymore. Is there some "as light API" as `get_mbed_com_port` was before ?

bridadan:
@jupe Totally agree 😄 It's absolutely on the list! As you can probably tell, @theotherjimmy and I have been in the process of cleaning up this repo since it previously had basically no testing or API boundaries.

To answer your question, the public API is basically just the one function that gives you all of the available data: https://github.com/ARMmbed/mbed-ls/blob/master/mbed_lstools/lstools_base.py#L110

However, as a general rule you can generally expect to call any function in [this class](https://github.com/ARMmbed/mbed-ls/blob/master/mbed_lstools/lstools_base.py#L53) that isn't marked as `deprecated` and doesn't start with an underscore (`_`).

I think you'll be especially interested in the `fs_interaction` and `platform_name_filters` options that we added there. It allows you to prevent reading the devices' filesystems to speed up the execution of mbed-ls. Hopefully that will allow to remove some of the custom checks y'all had to do in the past for speed reasons (not sure if y'all still do this though).

I think you'll be pretty happy with all of the testing that's been added (shoutout to @theotherjimmy for setting up all of the mocks) https://github.com/ARMmbed/mbed-ls/tree/master/test. We used to not test the OSX or Windows logic at all, so its a huge improvement. We're at 74% coverage and climbing! There's a lot of deprecated code that will be pulled out in the next major release that'll bump up that number 😄

[...]

PR #280, opened 2017-12-12
Replace name filter with function

bridadan:
This was written in response to the issues brought up in #277.

This replaces the `platform_name_filter` option with a `filter_function`. This now allows you to filter mbed devices by any available data point, not just the platform name.

This is _technically_ a breaking API change, but the filter option was only recently added and I don't think it has wide spread adoption.

FYI @jupe @theotherjimmy

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -109,14 +109,14 @@ class MbedLsToolsBase(object):
 
     def list_mbeds(
             self, fs_interaction=FSInteraction.BeforeFilter,
-            platform_name_filters=[], unique_names=False,
+            filter_function=None, unique_names=False,
             read_details_txt=False):
         """ List details of connected devices
         @return Returns list of structures with detailed info about each mbed
         @param fs_interaction A member of the FSInteraction class that picks the
           trade of between quality of service and speed
-        @param platform_name_filters A series of regular expressions to filter
-          filter targets by
+        @param filter_function Function that is passed each mbed candidate,
+          should return True if it should be included in the result
         @param unique_names A boolean controlling the presence of the
           'platform_unique_name' member of the output dict
         @param read_details_text A boolean controlling the presense of the
@@ -128,8 +128,6 @@ class MbedLsToolsBase(object):
         candidates = list(self.find_candidates())
         logger.debug("Candidates for display %r", candidates)
         result = []
-        platform_name_matcher = re.compile("|".join("({})".format(pf) for pf
-                                                    in platform_name_filters))
         for device in candidates:
             if  ((not device['mount_point'] or
                   not self.mount_point_ready(device['mount_point'])) and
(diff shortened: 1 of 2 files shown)


mbed_lstools/lstools_base.py
```diff
@@ -109,14 +109,14 @@ def list_mbeds_ext(self):
 
     def list_mbeds(
             self, fs_interaction=FSInteraction.BeforeFilter,
-            platform_name_filters=[], unique_names=False,
+            filter_function=None, unique_names=False,
```
jupe: could we get also public API tests? :) anyway, this looks better approach 👍

mbed_lstools/lstools_base.py
```diff
@@ -167,36 +165,38 @@ def list_mbeds(
...
         """Filter device before touching the file system of the device.
         Said another way: Touch the file system after filtering
         """
-        plat_name = self.plat_db.get(device['target_id_usb_id'][0:4])
-        if plat_name and not pn.match(plat_name):
-            return None
-        else:
+        device['target_id'] = device['target_id_usb_id']
+        device['target_id_mbed_htm'] = None
+        device['platform_name'] = self.plat_db.get(device['target_id'][0:4])
```
jupe: hey wht this give us when device is in bootloader mode now?

mbed_lstools/lstools_base.py
jupe: could be check that filter_function is really function when given

test/mbedls_toolsbase.py
```diff
@@ -176,16 +181,24 @@ def test_fs_after(self):
...
+            ret = self.base._fs_after_id_check(deepcopy(device), filter)
             self.assertIsNotNone(ret)
             self.assertEqual(ret['target_id'], new_device_id)
             _read_htm.assert_called_with(device['mount_point'])
 
             _read_htm.reset_mock()
 
-            filter_out = re.compile("NOT-K64F")
-            ret = self.base._fs_after_id_check(device, filter_out)
+            filter_in = lambda m: m['target_id'] == device['target_id_usb_id']
```
jupe: Propose to add this as example for public api doc string

mbed_lstools/lstools_base.py
```diff
@@ -167,36 +165,38 @@ def list_mbeds(
...
-    def _fs_after_id_check(self, device, pn):
+    def _fs_after_id_check(self, device, filter_function):
         """Filter device before touching the file system of the device.
         Said another way: Touch the file system after filtering
         """
-        plat_name = self.plat_db.get(device['target_id_usb_id'][0:4])
-        if plat_name and not pn.match(plat_name):
-            return None
-        else:
+        device['target_id'] = device['target_id_usb_id']
```
jupe: could you add “host_id” here as well, idea was to start using that as unique id because when daplink is in bootloader mode we have have only that... (could be in follow up pr also)

mbed_lstools/lstools_base.py
bridadan: Yeah, probably wouldn't hurt, that can be added in another PR.

test/mbedls_toolsbase.py
bridadan: Sure, can do

bridadan:
Added the example to the doc string 👍

mbed_lstools/lstools_base.py
jupe: yeah it belongs to own PR, just wanted to mention it :)

mbed_lstools/lstools_base.py
jupe: so API test for this is coming later, @theotherjimmy @bridadan ?

PR #282, opened 2017-12-14
Add ublox ODIN-W2 MTB

ashok-rao:
0450 = board ID assigned to new ublox ODIN W2 MTB.


ashok-rao:
@theotherjimmy  @bridadan .. Could you please review & merge if all OK? Thanks!

bridadan:
@ashok-rao we always do 😄 But if there's a timeline that you need this by please let us know so we can act accordingly. Thanks!

ashok-rao:
Thanks @bridadan . There won't be a platform page for these for now. So, I've not updated the slug. Name is now updated in the DB.

ashok-rao:
Thanks @bridadan . Will await the merge!

PR #287, opened 2018-01-09
Add MXCHIP EMW3166 MTB

ashok-rao:
0451 = board ID assigned to new MXCHIP EMW3166 MTB.
PE Platform database also updated. Target doesn't have a platform page, so slug field not updated.

@bridadan  @theotherjimmy .. could you please review / merge if all OK ? thanks!


theotherjimmy:
@bridadan I would appreciate it if you could double check the platform db.

PR #292, opened 2018-01-18
Add MTB MTS_Dragonfly

ashok-rao:
0454 = board ID assigned to MTB MTS dragonfly.
PE Platform database also updated. Target doesn't have a platform page, so slug field not updated.

@bridadan  @theotherjimmy .. could you please review / merge if all OK? Thanks!


ashok-rao:
@theotherjimmy .. could you please merge this in? We need it for the CI.. Thanks!

Issue #283, opened 2017-12-18
Detect boards with JLink firmware


sg-:
Some boards need to use JLink firmware (only one UART block) and the UART is used by the application. Has anyone investigated if JLink follows a standard pattern for identifying the board type?

@marhil01

bridadan:
I can speak at least for Windows this should be quite doable, the infrastructure is there to detect the HID endpoint we just currently aren't doing anything with it.

For Linux I think we need to do a little more investigation but I would imagine it shouldn't be too difficult. Same goes for Mac OS.

What information do you need in order to utilize the JLink/HID/Debug endpoint?

jupe:
We need at least platform name (seems that mbedls gives incorrectly this), target id (it gives something already, is it correct one?), mount point (probably ok already) and serial port. Basic things at least. Would be nice to have also jlink version etc but we can start with basic information. And support for linux is needed in our case.

bridadan:
> We need at least platform name (seems that mbedls gives incorrectly this), target id (it gives something already, is it correct one?)

Can you provide some debug output please (`mbedls --debug`)? We're working blind at this point.

Also please provide the normal info: target name, JLink version, Host OS platform, etc. These are all necessary so we can fix bugs/add features!

juhhov:
target: nRF52
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Raspbian
Description:	Raspbian GNU/Linux 8.0 (jessie)
Release:	8.0
Codename:	jessie
$ mbedls --version
1.3.5
$ mbedls -dj
DEBUG:mbedls.main:mbed-ls ver. 1.3.5
DEBUG:mbedls.main:host: ('posix', 'Linux', '4.4.26-v7+', '#915 SMP Thu Oct 20 17:08:44 BST 2016', 'linux2')
DEBUG:mbedls.lstools_linux:Converting device list ['/dev/disk/by-id/usb-SEGGER_MSD_Volume_000682245000-0:0', '/dev/disk/by-id/mmc-SD32G_0x0028012d-part2', '/dev/disk/by-id/mmc-SD32G_0x0028012d-part1', '/dev/disk/by-id/mmc-SD32G_0x0028012d']
DEBUG:mbedls.lstools_linux:Found disk devices by id {'000682245000': '/dev/sda'}
DEBUG:mbedls.lstools_linux:Converting device list ['/dev/serial/by-id/usb-SEGGER_J-Link_000682245000-if00']
[...]

PR #299, opened 2018-01-30
Jlink and other fw detection

bridadan:
This is a proposal to address #283.

One thing that needs more investigation is how to properly identify JLink devices. Currently I'm assuming that the target id is unique and corresponds with the `platform_name`, just like DAPLink. I need to check other platforms running JLink to verify this.

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -204,9 +204,12 @@ class MbedLsToolsBase(object):
             return
 
         device['device_type'] = self._detect_device_type(device['mount_point'])
-
         device['target_id'] = device['target_id_usb_id']
-        self._update_device_details(device, include_extra_info)
+
+        {
+            'daplink': self._update_device_details_daplink,
+            'jlink': self._update_device_details_jlink
+        }[device['device_type']](device, include_extra_info)
 
     def _detect_device_type(self, mount_point):
         """ Returns a string of the device type
@@ -216,17 +219,6 @@ class MbedLsToolsBase(object):
         files = [f.lower() for f in os.listdir(mount_point)]
         return 'jlink' if 'segger.html' in files else 'daplink'
 
-    def _update_device_details(self, device, include_extra_info):
-        """ Updates device dict with information from the filesystem
-            @param device Device dictionary to update. Should contain a valid
-                'mount_point' value
-            @param include_extra_info Controls how much information (and by extension,
-                the performance) is returned from the filesystem
-        """
-        {
-            'daplink': self._update_device_details_daplink,
-            'jlink': self._update_device_details_jlink
-        }[device['device_type']](device, include_extra_info)
 
     def _update_device_details_daplink(self, device, include_extra_info):
         self._update_device_from_htm(device)
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -98,19 +98,19 @@ class MbedLsToolsBase(object):
 
     @deprecated("Functionality has been moved into 'list_mbeds'. "
                 "Please use list_mbeds with 'unique_names=True' and "
-                "'read_details_txt=True'")
+                "'include_extra_info=True'")
     def list_mbeds_ext(self):
         """! Function adds extra information for each mbed device
         @return Returns list of mbed devices plus extended data like 'platform_name_unique'
         @details Get information about mbeds with extended parameters/info included
         """
 
-        return self.list_mbeds(unique_names=True, read_details_txt=True)
+        return self.list_mbeds(unique_names=True, include_extra_info=True)
 
     def list_mbeds(
             self, fs_interaction=FSInteraction.BeforeFilter,
             filter_function=None, unique_names=False,
-            read_details_txt=False):
+            include_extra_info=False):
         """ List details of connected devices
         @return Returns list of structures with detailed info about each mbed
         @param fs_interaction A member of the FSInteraction class that picks the
@@ -120,8 +120,8 @@ class MbedLsToolsBase(object):
           Ex. mbeds = list_mbeds(filter_function=lambda m: m['platform_name'] == 'K64F')
         @param unique_names A boolean controlling the presence of the
           'platform_unique_name' member of the output dict
-        @param read_details_text A boolean controlling the presense of the
-          output dict attributes read from "DETAILS.TXT"
+        @param include_extra_info A boolean controlling the presense of the
+          output dict attributes read from other files present on the 'mount_point'
         @details Function returns list of dictionaries with mbed attributes 'mount_point', TargetID name etc.
         Function returns mbed list with platform names if possible
         """
@@ -143,7 +143,7 @@ class MbedLsToolsBase(object):
                     FSInteraction.BeforeFilter: self._fs_before_id_check,
                     FSInteraction.AfterFilter: self._fs_after_id_check,
                     FSInteraction.Never: self._fs_never
-                }[fs_interaction](device, filter_function, read_details_txt)
+                }[fs_interaction](device, filter_function, include_extra_info)
                 if maybe_device:
                     if unique_names:
                         name = device['platform_name']
@@ -151,6 +151,7 @@ class MbedLsToolsBase(object):
                         platform_count[name] += 1
                         device['platform_name_unique'] = (
                             "%s[%d]" % (name, platform_count[name]))
+
                     try:
                         device.update(self.retarget_data[device['target_id']])
                         logger.debug("retargeting %s with %r",
@@ -162,7 +163,7 @@ class MbedLsToolsBase(object):
 
         return result
 
-    def _fs_never(self, device, filter_function, read_details_txt):
+    def _fs_never(self, device, filter_function, include_extra_info):
         """Filter device without touching the file system of the device"""
         device['target_id'] = device['target_id_usb_id']
         device['target_id_mbed_htm'] = None
@@ -172,23 +173,18 @@ class MbedLsToolsBase(object):
         else:
             return None
 
-    def _fs_before_id_check(self, device, filter_function, read_details_txt):
+    def _fs_before_id_check(self, device, filter_function, include_extra_info):
         """Filter device after touching the file system of the device.
         Said another way: Touch the file system before filtering
         """
-        self._update_device_from_htm(device)
-
-        if read_details_txt:
-            details_txt = self._details_txt(device['mount_point']) or {}
-            device.update({"daplink_%s" % f.lower().replace(' ', '_'): v
-                           for f, v in details_txt.items()})
-
+        self._update_device_from_fs(device, include_extra_info)
         if not filter_function or filter_function(device):
             return device
         else:
             return None
 
-    def _fs_after_id_check(self, device, filter_function, read_details_txt):
+    def _fs_after_id_check(self, device, filter_function, include_extra_info):
+
         """Filter device before touching the file system of the device.
         Said another way: Touch the file system after filtering
         """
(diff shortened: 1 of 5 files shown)


mbed_lstools/platform_database.py
```diff
@@ -47,7 +47,13 @@
     u'0003': u'LPC2368',
     u'0004': u'LPC2368',
     u'0005': u'LPC2368',
-    u'0006': u'LPC2368',
+    u'0006': {
+        u'default': 'daplink',
```
jupe: when this default is needed?

bridadan: If someone invokes the `platform_db.get` api without specifying a `device_type`, then we need to know which platform mapping we need to use. I'm open to restructuring that bit if needed, it was just a quick pass at it.

bridadan:
Bad news: I've tried flashing a KL25Z, a KL27Z, and a KL43Z using the JLink binaries found here: https://www.segger.com/downloads/jlink/#JLinkOpenSDABoardSpecificFirmwares

All of them have a usb target ID of `000000123456`. So it doesn't like this is reliable for detection. It also means that they won't be unique in CI if multiple are plugged into the same machine. This may not be an issue for the NRF52_DK Jlink build as it seems to be more unique.

But the point is my modifications to the platform database won't work. Instead the DB needs to have some significant changes to it most likely.

bridadan:
This actually works! But it needs some polish, some comments, and most importantly tests.

@theotherjimmy How do you feel about the platform database changes? I don't _love_ them but they work.

bridadan:
I got more tests to write

bridadan:
I'm actually starting to become reasonably happy with this!

A few things to note:

- I have renamed the parameter `read_details_txt` to `include_extra_info` in the top level API, which is a breaking change
    - This rename was because not all firmware store the extra info in a file called `DETAILS.TXT`
    - Instead of renaming the parameter I could alias the parameter so both of them did the same thing. A deprecation warning could be issued for the `read_details_txt` parameter when used.
        - If that sounds ok to you @theotherjimmy then I'll go ahead and make that change
- If it can't determine that the device is using Jlink (`segger.html` file is present), then it will assume it is a daplink device. This is fairly safe but could be misleading.
    - Ex. ST-Link is typically completely compatible with daplink, but it maybe useful to try and detect if it is an st-link device for management purposes.

theotherjimmy:
@bridadan 
> Instead of renaming the parameter I could alias the parameter so both of them did the same thing. A deprecation warning could be issued for the read_details_txt parameter when used. 

👍 

> If it can't determine that the device is using Jlink (segger.html file is present), then it will assume it is a daplink device. This is fairly safe but could be misleading. 

As usual, it should respect the FsInteraction.<whatever> options.

bridadan:
@theotherjimmy 

> > If it can't determine that the device is using Jlink (segger.html file is present), then it will assume it is a daplink device. This is fairly safe but could be misleading.

> As usual, it should respect the FsInteraction. options.

It does do this currently. But in the case where the files don't match a JLink device, it will default to daplink.

I should also mention this **changes the format of the platform database**. Previously it was this:

```
{
    "0000": "name"
}
```

Now it's this:

```
{
    "daplink": {
        "0000": "name"
    },
    "jlink": ...
}
```

I believe this should work on systems that already have a platform db on the filesystem, but it does change how you interact with it when not using daplink devices.

theotherjimmy:
@bridadan The changes to the platform DB should be fine. I forgot that it already gracefully upgrades.

bridadan:
@theotherjimmy at this point I'm pretty much done with the work I had in mind, let me know if you see any issues or want anything changed

theotherjimmy:
Given that you don't use the `include_extra_info` flag in the JLink implementation, it has exactly the semantics of `read_details_text`. I see no point it adding the rename **yet**. Until it means something for JLink, I think the name change is unneeded churn.

bridadan:
Agreed, I've reverted the change to the parameter.

theotherjimmy:
Awesome. Thanks!

juhhov:
This seems to work :+1: 

One thing to consider. We would like to use the mbedls platform_name for JLinkExe device parameter.
For nrf52 current platform_name NRF52_DK is not recognized. Here is the complete list of JLinkExe supported boards https://www.segger.com/downloads/supported-devices.php , for nrf52 boards select Nordic Semi from the list at the top.
Would it be possible to use platform_name from that list? Thoughts @bridadan , @jupe ?

bridadan:
@juhhov, that should definitely be possible, although we'll have to return it as a separate data value instead of the `platform_name` for compatibility reasons with the rest of the Mbed OS tools. Probably something like `jlink_platform_name`.

In the case of the NRF52_DK, what is the correct device name from that list? There seems to be four options: nRF52810_xxAA, nRF52832_xxAA, nRF52832_xxAB, and nRF52840_xxAA.

simtind:
It's an nRF52832_xxaa, ab is a device with smaller memory that has no separate dev kit.

jupe:
@simtind do you know if there is way to detect board/chip model from dev kit using usb port ? Which can be used for `JLinkExe --device <here>`.

simtind:
Nothing for the general case. What you can do is connect as some generic cpu (Cortex-m0 fex.) and check the cpuid register. That should tell you at least what core you're working on. Then you can reconnect as the correct core, which should give you access to memory.

For nRF devices we base the Jlink snr on the dev kit as you've guessed in the issue.
The two first numbers (in decimal) are always 68, the third number is 0 for nRF51-Dongle, 1 for nRF51-DK, 2 for nRF52-DK and 3 for nrF52840-PDK.

bridadan:
Ok, rebased and added jlink information to the platform db. Probably need to add a few more tests for some missing coverage.

bridadan:
Coveralls is taking its sweet time but I added the coverage back 👍

juhhov:
@bridadan this works for us. Could this move forward?

theotherjimmy:
@bridadan Could you rebase? I'll review it right now, assuming that little has to change when rebasing.

mbed_lstools/lstools_base.py
```diff
@@ -110,7 +110,7 @@ def list_mbeds_ext(self):
     def list_mbeds(
             self, fs_interaction=FSInteraction.BeforeFilter,
             filter_function=None, unique_names=False,
-            read_details_txt=False):
+            read_details_txt=None):
```
theotherjimmy: Could you revert this change? It seems like a "style-like" thing to me. I don't think this code explicitly looks for `None` instead of any falsey value

mbed_lstools/lstools_base.py
```diff
@@ -151,6 +151,7 @@ def list_mbeds(
                         platform_count[name] += 1
                         device['platform_name_unique'] = (
                             "%s[%d]" % (name, platform_count[name]))
+
```
theotherjimmy: Could you remove this formatting change?

mbed_lstools/lstools_base.py
```diff
@@ -176,37 +179,109 @@ def _fs_before_id_check(self, device, filter_function, read_details_txt):
...
 
+        device['target_id'] = device['target_id_usb_id']
+        self._update_device_from_fs(device, read_details_txt)
         if not filter_function or filter_function(device):
             return device
         else:
             return None
 
     def _fs_after_id_check(self, device, filter_function, read_details_txt):
+
```
theotherjimmy: Could you remove this formatting change?

mbed_lstools/lstools_base.py
```diff
@@ -176,37 +179,109 @@ def _fs_before_id_check(self, device, filter_function, read_details_txt):
...
+        if device['target_id']:
+            platform_data = self.plat_db.get(device['target_id'][0:4],
+                                             device_type='daplink',
+                                             verbose_data=True)
+            device.update(platform_data or {'platform_name': None})
+        else:
+            device['platform_name'] = None
+
+    def _update_device_details_jlink(self, device, read_details_txt):
```
theotherjimmy: Could you change `read_details_txt` to `_` to indicate that you are not using that parameter?

mbed_lstools/lstools_base.py
```diff
@@ -176,37 +179,109 @@ def _fs_before_id_check(self, device, filter_function, read_details_txt):
...
+        """ Updates the jlink-specific device information based on files from its 'mount_point'
+            @param device Dictionary containing device information
+            @param read_details_txt A boolean controlling the presense of the
+              output dict attributes read from other files present on the 'mount_point'
+        """
+        files = os.listdir(device['mount_point'])
```
theotherjimmy: Could we list the directory only once some how? We could do the listdir outside this function and `_detect_device_type` and pass it into both. I think that this may have an impact on performance (we're somewhat performance critical to RAAS, IIRC) and we want to avoid making things any slower than they have to be.

mbed_lstools/platform_database.py
```diff
@@ -42,228 +42,248 @@
...
+        u'C033': u'UBLOX_C030_R410M',
+        u'C034': u'UBLOX_C030_S200',
+        u'C035': u'UBLOX_C030_R3121',
+        u'FFFF': u'K20 BOOTLOADER',
+        u'RIOT': u'RIOT',
+    },
+    u'jlink': {
+        u'X349858SLYN': {
+            u'platform_name': u'NRF52_DK',
+            u'jlink_device_name': u'nRF52832_xxaa'
```
theotherjimmy: Why do we need this data? Is it important to keep?

bridadan: This is needed for the reason specified in this comment: https://github.com/ARMmbed/mbed-ls/pull/299#issuecomment-363059366

mbed_lstools/lstools_base.py
bridadan: Oops good catch. That was from when I had initially renamed the parameter.

bridadan:
Rebased!

jupe:
so how json result looks like now for jlink based board?

bridadan:
@jupe Like so:

```json
$ mbedls -j
[
    {
        "device_type": "jlink",
        "jlink_device_name": "nRF52832_xxaa",
        "mount_point": "D:",
        "platform_name": "NRF52_DK",
        "platform_name_unique": "NRF52_DK[0]",
        "serial_port": "COM32",
        "target_id": "000682961642",
        "target_id_usb_id": "000682961642",
        "url": "http://www.nordicsemi.com/X349858SLYN"
    }
]
```

jupe:
looks good. Is  `device_type` property exists allways, and `daplink` for daplink based boards?

bridadan:
Yup!

```json
$ mbedls -j
[
    {
        "daplink_auto_reset": "0",
        "daplink_automation_allowed": "1",
        "daplink_bootloader_crc": "0xa65218eb",
        "daplink_bootloader_version": "0242",
        "daplink_daplink_mode": "Interface",
        "daplink_git_sha": "67f8727a030bcc585e982d899fb6382db56d673b",
        "daplink_hic_id": "97969900",
        "daplink_interface_crc": "0xe4422294",
        "daplink_interface_version": "0244",
        "daplink_local_mods": "0",
        "daplink_overflow_detection": "1",
        "daplink_remount_count": "0",
        "daplink_unique_id": "0240000032044e4500257009997b00386781000097969900",
        "daplink_usb_interfaces": "MSD, CDC, HID",
        "daplink_version": "0244",
        "device_type": "daplink",
        "mount_point": "D:",
        "platform_name": "K64F",
        "platform_name_unique": "K64F[0]",
        "serial_port": "COM18",
        "target_id": "0240000032044e4500257009997b00386781000097969900",
        "target_id_mbed_htm": "0240000032044e4500257009997b00386781000097969900",
        "target_id_usb_id": "0240000032044e4500257009997b00386781000097969900"
    }
]
```

theotherjimmy:
@bridadan Are you happy with this implementation?

bridadan:
Yup!

PR #304, opened 2018-02-07
Add DISCO_F303VC

jeromecoutant:
Add DISCO_F303VC

http://www.st.com/en/evaluation-tools/stm32f3discovery.html


jeromecoutant:
I can set F303C

Then I will update MBED-OS side accordingly

bridadan:
@jeromecoutant Your product page lists it as STM32F303**V**C though, won't that cause confusion?

PR #311, opened 2018-02-28
Default to NullHandler when logging not configured

### LINKED ISSUE #302: No logging configuration if create used
If the create-method in mbed_lstools main is used to create the toolset, the logging module never receives any configuration so nothing ever gets logged. Is it assumed that the application calling create would configure the logging itself or should the start_logging  method be automatically called in create?

theotherjimmy:
Resolves #302


theotherjimmy:
@bridadan Could you review?

theotherjimmy:
I tried it, nothing changed.

PR #312, opened 2018-02-28
Document logging behavior

### LINKED ISSUE #302: No logging configuration if create used

theotherjimmy:
Resolves #302 (Docmumentation bits)

@AnotherButler Could you edit?

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,6 +286,10 @@ Overrides the `platform_name` for any platform with a `target_id` that starts wi
 
 If set to `'+'`, the mocked platform is enabled. If `'-'`, the mocked platform is disabled.
 
+## Logging
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -286,6 +286,10 @@ Overrides the `platform_name` for any platform with a `target_id` that starts wi
 
 If set to `'+'`, the mocked platform is enabled. If `'-'`, the mocked platform is disabled.
 
+## Logging
+
+Mbed LS uses the python `logging` module for all of its logging needs. Mbed LS uses the logger `"mbedls"` as its root, and all other loggers start with `"mbedls."`. Configuring the python root logger automatically redirects all of the logging done by Mbed LS to the configured endpoint. When using the python API, configure logging, such as by calling `logging.basicConfig()`.
```
AnotherButler: - Please capitalize "Python" all three times. 
- Please change "all the logging done by Mbed LS" to "all the logging Mbed LS does" for active voice.

PR #315, opened 2018-03-14
Add NUCLEO_WB55RG

jeromecoutant:
Add NUCLEO_WB55RG

Thx


jeromecoutant:
Maybe we can add it in the next release ?

PR #316, opened 2018-03-23
add MTB_RAK811

ithinuel:
0457 = board ID assigned to module test board for the RAK-811.


ashok-rao:
PE platform database updated with board ID. Target doesn't have a platform page on the website yet, so slug field not updated.
@theotherjimmy  @bridadan .. could you please review / merge? Thanks.

PR #318, opened 2018-03-29
Add UNO_81C/UNO_81AM/UNO_81A to the list of boards

RdaZhongyao:
New targets from RDA Microelectronics, @Eric-Yang-China  @Ronny-Liu


theotherjimmy:
Waiting on CI again. Just a few min.

PR #328, opened 2018-04-17
Add new Maxim targets

bentcooke:
Adding the following Maxim targets and IDs:
        0416 MAX32625PICO
        0417 MAX32630MBED
        0418 MAX32620FTHR
        0419 MAX35103EVKIT2
        0420 MAX32630HSP3

@theotherjimmy, could you please review?


bentcooke:
Yes, those targets do not have platform pages yet.

PR #329, opened 2018-04-17
Windows detection of non composite devices (ex bootloader mode)

bridadan:
This fixes an issue where devices in bootloader mode are no longer shown in Windows.

This is a fairly extensive detection strategy change, so this PR needs to go through some more testing, especially with actual hardware.

FYI @c1728p9

### DIFF
--- a/mbed_lstools/windows.py
+++ b/mbed_lstools/windows.py
@@ -38,7 +38,7 @@ else:
 
 
 MAX_COMPOSITE_DEVICE_SUBDEVICES = 5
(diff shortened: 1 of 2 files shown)


bridadan:
Oh also, could use some more tests

bridadan:
I've tested with a number of actual pieces of hardware and the new detection seems to work well. Next step is to add a few more tests to get the coverage back up.

bridadan:
Tests added!

PR #335, opened 2018-04-26
Patch edge case for devices with unreadable files

bridadan:
If a device has a valid mount point but the files on present are not readable, the device should not be listed in the results (unless `list_unmounted` is `True`). As found by @jupe, we were handling this error farther down in the software stack than we should have, and thus the error was being masked/handled to quickly. This led to devices with unreadable files showing up in the list.

This PR adds a test case and a fix for this issue.

cc @theotherjimmy

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -217,7 +217,7 @@ class MbedLsToolsBase(object):
                 'daplink': self._update_device_details_daplink,
                 'jlink': self._update_device_details_jlink
             }[device['device_type']](device, read_details_txt, directory_entries)
-        except OSError as e:
+        except (OSError, IOError) as e:
             logger.warning(
                 'Marking device with mount point "%s" as unmounted due to the '
                 'following error: %s', device['mount_point'], e)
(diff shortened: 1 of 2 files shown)


theotherjimmy:
Wow 80% :champagne:

bridadan:
It increased because I removed code, which of course is a good thing! 😁

jupe:
Any change to get patch release with this fix soon ?

jupe:
Do we get patch release soon?

jupe:
This actually leads to another fatal issue :/ . See:
```
$ mbedls -u
WARNING:mbedls.platform_database:Duplicate platform ids found: daplink.0461 daplink.0462 daplink.0457, ignoring the definitions from /home/user/.local/share/mbedls/platforms.json
WARNING:mbedls.lstools_base:Marking device with mount point "/mnt/platform-3f980000_usb-usb-0_1_4_1_0-scsi-0_0_0_0" as unmounted due to the following error: [Errno 2] No such file or directory: '/mnt/platform-3f980000_usb-usb-0_1_4_1_0-scsi-0_0_0_0/mbed.htm'
| K66F          | K66F[0]              | unknown     | /dev/ttyACM0 | 0311000033514e45000d5015846c00176cb1000097969900 | unknown         |
```

bridadan:
Yes we saw this yesterday, we'll get a patch up soon. In the meantime you may need to pin 1.4.4 in your automated systems.

PR #338, opened 2018-05-08
Fix detection of NRF51_DK on windows.

### LINKED ISSUE #337: NRF51_DK incorrect detection on Windows 10 x64
Hi, I'm trying to run `mbed test` on my windows 10 development machine against a standard NRF51_DK board.
In 1.4.4 and 1.4.3 it's being detected as:
```
+---------------+----------------------+-------------+-------------+--------------+-----------------+
| platform_name | platform_name_unique | mount_point | serial_port | target_id    | daplink_version |
+---------------+----------------------+-------------+-------------+--------------+-----------------+
| LPC2368       | LPC2368[0]           | K:          | COM3        | 000681078661 | unknown         |
+---------------+----------------------+-------------+-------------+--------------+-----------------+
```
That target id is also visible in device manager under the Parent property as `USB\VID_1366&PID_1015\000681078661`
In code it's pulled from the volume_string in `find_candidates()`
Where volume_string = `u'_??_USBSTOR#Disk&Ven_SEGGER&Prod_MSD_Volume&Rev_1.00#8&2ee71873&0&000681078661&0#{53f56307-b6bf-11d0-94f2-00a0c91efb8b}'`

I've traced the detection to looking at the first 4 characters in target_id and mapping that in the json database to "0006": "LPC2368"

mbed_lstools/lstools_base.py:143
`platform_data = self.plat_db.get(device['target_id_usb_id'][0:4], verbose_data=True)`

In 1.4.2 it comes up as unknown. It's unknown all the way back to 1.3.3, 1.3.2 doesn't run due too logging import error then 1.3.1 doesn't detect any device.

1.2.16 detects my board perfectly, though the current mbed-greentea won't run against this old a version of mbedls.
```
+---------------+----------------------+-------------+-------------+--------------------------------------------------+-----------------+
| platform_name | platform_name_unique | mount_point | serial_port | target_id                                        | daplink_version |
+---------------+----------------------+-------------+-------------+--------------------------------------------------+-----------------+
| NRF51_DK      | NRF51_DK[0]          | K:          | COM3        | 1100021852333120343935373230322031323033AFD4DFD8 | unknown         |
+---------------+----------------------+-------------+-------------+--------------------------------------------------+-----------------+
```
It's pulling target_id from the htm which is the correct ID, [0:4] == 1100

andrewleech:
Should also work for other jlink based boards which also have mbed.htm on the mass storage interface.

Closes #337

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -140,7 +140,7 @@ class MbedLsToolsBase(object):
                         "Use the '-u' flag to include it in the list.",
                         device['target_id_usb_id'])
             else:
-                platform_data = self.plat_db.get(device['target_id_usb_id'][0:4], verbose_data=True)
+                platform_data = self.plat_db.get(device['target_id_usb_id'], verbose_data=True)
                 device.update(platform_data or {"platform_name": None})
                 maybe_device = {
                     FSInteraction.BeforeFilter: self._fs_before_id_check,
@@ -248,7 +248,7 @@ class MbedLsToolsBase(object):
                            for f, v in details_txt.items()})
 
         if device['target_id']:
-            platform_data = self.plat_db.get(device['target_id'][0:4],
+            platform_data = self.plat_db.get(device['target_id'],
                                              device_type='daplink',
                                              verbose_data=True)
             if platform_data:
@@ -263,8 +263,14 @@ class MbedLsToolsBase(object):
         """
         lower_case_map = {e.lower(): e for e in directory_entries}
 
-        if 'board.html' in lower_case_map:
+        device_type = 'jlink'
+
+        if 'mbed.htm' in lower_case_map:
+            board_file_key = 'mbed.htm'
+            device_type = 'daplink'  # If mbed.htm exists we can use main device db section
+        elif 'board.html' in lower_case_map:
             board_file_key = 'board.html'
+
         elif 'user guide.html' in lower_case_map:
             board_file_key = 'user guide.html'
         else:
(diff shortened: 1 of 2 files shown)


bridadan:
Hi @andrewleech, thanks so much for the detailed issue report in #337, you made it a breeze to reproduce the issue!

And thanks for being proactive and submitting a fix! Your fix will definitely work, but from my point of view the preferred fix would be just your addition to the `platform_database` here: https://github.com/andrewleech/mbed-ls/blob/273c5011bce3ff9fe3d53d6aba6ef602fa4aa948/mbed_lstools/platform_database.py#L290-L293

The reasoning here is we really try to reduce the number of paths the detection process can take. JLink firmware in particular is difficult because the actual files on the different boards differs so much. Some boards have the `mbed.htm` like you noticed, but a lot don't.

Another reason we want to ensure we're using the `jlink` entry in the `platform_database` is because we rely on the `jlink_device_name` present in the db for other tools. If we used the `daplink` entry, we'd be missing that information.

Hopefully that all makes sense, let me know if you have any follow up questions or suggestions!

andrewleech:
Yeah ok, at the time I thought I should fix detection of the official mbed target id but I guess that doesn't really matter as long after the used board id is unique.

bridadan:
@andrewleech That's a good point about the uniqueness of the ID. If that turns out to be an issue than we'll probably need to revisit this issue and making use of the `mbed.htm` ID. But for now I'd go for the simple platform_database entry if possible.

andrewleech:
The only real need for the target id is to be unique on a particular pc (and it's a usb id so that should be safe) and to be repeatable on said machine (critical for headless / ci build machine). I can't speak to the repeatability personally but yes, I'm happy to stick with the current implementation until we find out otherwise.

I'll make a new PR with the db update fix, we can abandon this PR.

FWIW I got my hands on an nRF52840-Preview-DK here (it's the new bigger nrf52), unfortunately its ID from `User Guide.html` is the same as the regular NRF52_DK so presumably will only match against code built for the original nRF52832. It does have an `MBED.htm` but it does _not_ have a link with an ID in it, just a link to mbed.com
I think we can chalk it up to being a preview dk, that will be stuff they didn't finish and hopefully a final release 840 dk would be better. But yes as you've said, we really can't rely on what the manf puts on the flash drive.

PR #340, opened 2018-05-09
Platform DB: Add NRF51_DK to jlink section.

andrewleech:
Adds the required details to the platform db as read from a nordic semiconductor nRF51 DK  PCA10028 V1.2.2

### DIFF
--- a/mbed_lstools/platform_database.py
+++ b/mbed_lstools/platform_database.py
@@ -287,6 +287,10 @@ DEFAULT_PLATFORM_DB = {
         u'RIOT': u'RIOT',
     },
     u'jlink': {
+        u'X729475D28G': {
+            u'platform_name': u'NRF51_DK',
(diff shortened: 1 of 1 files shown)


mbed_lstools/platform_database.py
```diff
@@ -287,6 +287,10 @@
         u'RIOT': u'RIOT',
     },
     u'jlink': {
+        u'X729475D28G': {
+            u'platform_name': u'NRF51_DK',
+            u'jlink_device_name': u'nRF51422_xxAC'
```
bridadan: Shouldn't the `jlink_device_name` be `nRF51822_xxAC`?

andrewleech: No, the DK actually has the 422 on it, which can run code written for a few of their chips.

Issue #251, opened 2017-10-20
Mock interface difficult to use and error prone


theotherjimmy:
Currently, to add a mock, you have to do some odd things:
```bash
mbedls -m "C006:VLBUNO51"
```
```bash
mbedls --mock="C006:VLBUNO51"
```
both add a mock for the target id `C006` and the platform name `VLBUNO51`.

While that my seem reasonable, it can also be done by:
```bash
mbedls --mock="+C006:VLBUNO51"
```
Which is a bit complicated.

Removing mocks are where things get nasty:
```bash
mbedls -m "-C006"
```
is in theory part of the syntax, but `-C006` is parsed as a flag.
```bash
mbedls --mock="-C006"
```
is the only working way to actually remove a mock.
The readme also gives the impression that:
```bash
mbedls --mock="!C006"
```
should work, but it does not. Instead, it gives you a lovely trackback:
```
Traceback (most recent call last):
  File "/usr/bin/mbedls", line 11, in <module>
    load_entry_point('mbed-ls==1.3.2', 'console_scripts', 'mbedls')()
  File "/usr/lib/python3.6/site-packages/mbed_lstools/main.py", line 258, in mbedls_main
    ret_code = args.command(mbeds, args)
  File "/usr/lib/python3.6/site-packages/mbed_lstools/main.py", line 127, in mock_platform
    mbeds.mock_manufacture_id(mid, 'dummy', oper=oper)
  File "/usr/lib/python3.6/site-packages/mbed_lstools/lstools_base.py", line 231, in mock_manufacture_id
    raise ValueError("oper can only be [+-]")
ValueError: oper can only be [+-]
```

A much better interface would be:
```bash
mbedls add-mock C006 VLBUNO51
```
for add and:
```bash
mbedls rm-mock C006
```
or
```bash
mbedls remove-mock C006
```
for mock removal.

@yennster @bridadan Does this cover our offline discussion?

yennster:
Looks good to me, the readme also gives the impression that 
```
mbedls --mock="!*"
```
should work to remove all mocks.

Which would be a useful feature if mocks gets re-worked.

theotherjimmy:
@yennster I think that you would have to do 
```
mbedls --mock="-*"
```
right now

theotherjimmy:
Any thoughts on my proposed new interface?

yennster:
@theotherjimmy 😢 What does the "!" do then?

yennster:
[...]

PR #341, opened 2018-05-09
Adding a --update command for pulling new target defintions from online

bridadan:
This is a very basic implementation of #119.

FYI @theotherjimmy @jupe 

**TODO**

- [x] Indicate success of update with error code
- [x] Throw exception on failed update (edit: using a return code)
- [x] Add doc string for public api
- [x] Document structure of JSON for plat_db
- [x] Add url argument for the update cli command
~~- [ ] Change the `--update` argument to a full subcommand (ex `mbedls update`)~~
        - Edit: holding off on this for a later PR
- [x] Add more log output during the update

### DIFF
--- a/mbed_lstools/main.py
+++ b/mbed_lstools/main.py
@@ -130,6 +130,9 @@ def mock_platform(mbeds, args):
         else:
             logger.error("Could not parse mock from token: '%s'", token)
 
+def update_from_web(mbeds, args):
+    mbeds.plat_db.update_from_web()
+
 def list_platforms(mbeds, args):
     print(mbeds.list_manufacture_ids())
 
@@ -204,6 +207,9 @@ def parse_cli(to_parse):
         '-m', '--mock', metavar='ID:NAME',
         help='substitute or create a target ID to platform name mapping used'
         'when invoking mbedls in the current directory')
+    commands.add_argument(
+        '-U', '--update', action='store_true',
+        help='update the platform database from os.mbed.com')
 
     parser.add_argument(
         '--skip-retarget', dest='skip_retarget', default=False,
@@ -221,6 +227,8 @@ def parse_cli(to_parse):
     args = parser.parse_args(to_parse)
     if args.mock:
         args.command = mock_platform
+    elif args.update:
+        args.command = update_from_web
     return args
 
 def start_logging():
--- a/mbed_lstools/main.py
+++ b/mbed_lstools/main.py
@@ -26,6 +26,7 @@ from collections import defaultdict
 
 # Make sure that any global generic setup is run
 from . import lstools_base
+from .platform_database import DEFAULT_UPDATE_URL
 
 import logging
 logger = logging.getLogger("mbedls.main")
@@ -131,7 +132,7 @@ def mock_platform(mbeds, args):
             logger.error("Could not parse mock from token: '%s'", token)
 
 def update_from_web(mbeds, args):
-    mbeds.plat_db.update_from_web()
+    return 0 if mbeds.plat_db.update_from_web(args.update_url) else 1
 
 def list_platforms(mbeds, args):
     print(mbeds.list_manufacture_ids())
@@ -208,8 +209,9 @@ def parse_cli(to_parse):
         help='substitute or create a target ID to platform name mapping used'
         'when invoking mbedls in the current directory')
     commands.add_argument(
-        '-U', '--update', action='store_true',
-        help='update the platform database from os.mbed.com')
+        '-U', '--update', nargs='?', default=False,
+        const=DEFAULT_UPDATE_URL, dest='update_url',
+        help='update the platform database from a URL (default: %s)' % (DEFAULT_UPDATE_URL))
 
     parser.add_argument(
         '--skip-retarget', dest='skip_retarget', default=False,
@@ -227,7 +229,7 @@ def parse_cli(to_parse):
     args = parser.parse_args(to_parse)
     if args.mock:
         args.command = mock_platform
-    elif args.update:
+    elif args.update_url:
         args.command = update_from_web
     return args
 
--- a/mbed_lstools/main.py
+++ b/mbed_lstools/main.py
@@ -26,7 +26,8 @@ from collections import defaultdict
 
 # Make sure that any global generic setup is run
 from . import lstools_base
-from .platform_database import DEFAULT_UPDATE_URL
+from .platform_database import DEFAULT_UPDATE_URL, RemotePlatformDataException,\
+                               DatabaseUpdateException
 
 import logging
 logger = logging.getLogger("mbedls.main")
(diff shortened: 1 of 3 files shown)


theotherjimmy:
:balloon: :tada: :birthday:

mbed_lstools/main.py
```diff
@@ -130,6 +130,9 @@ def mock_platform(mbeds, args):
         else:
             logger.error("Could not parse mock from token: '%s'", token)
 
+def update_from_web(mbeds, args):
+    mbeds.plat_db.update_from_web()
```
jupe: would be good to indicate if update success or not as a retcode

mbed_lstools/platform_database.py
```diff
@@ -478,3 +479,16 @@ def remove(self, id, permanent=False, device_type='daplink', verbose_data=False)
                     self._update_db()
 
                 return _modify_data_format(removed, verbose_data)
+
+    def update_from_web(self, url='https://os.mbed.com/api/v3/platforms/?format=json'):
```
jupe: is this public API ? this could throw some specific exception when updating fails for some reason

test/platform_database.py
```diff
@@ -149,6 +149,26 @@ def test_simplify_verbose_data(self):
         self.assertEqual(self.pdb.get('1337', verbose_data=True), platform_data)
         self.assertEqual(self.pdb.get('1337'), platform_data['platform_name'])
 
+    def test_update_from_web(self):
+        with patch("requests.get") as _get:
+            self.assertEqual(self.pdb.get('1337'), None)
+            _result = MagicMock()
+            _result.configure_mock(**{
+                'json.return_value': [
+                    {
```
jupe: maybe could be good to document this json format

mbed_lstools/main.py
```diff
@@ -204,6 +207,9 @@ def parse_cli(to_parse):
         '-m', '--mock', metavar='ID:NAME',
         help='substitute or create a target ID to platform name mapping used'
         'when invoking mbedls in the current directory')
+    commands.add_argument(
```
jupe: maybe here could be also optional update url argument that is used to fetch updates - with default value which are visible in help..

mbed_lstools/main.py
theotherjimmy: Agreed

mbed_lstools/platform_database.py
theotherjimmy: @bridadan This is a public API, and it should have a docstring, yeah?

test/platform_database.py
theotherjimmy: Good point. FYI, this is a test.

mbed_lstools/main.py
theotherjimmy: Definitely. Very good spot!

mbed_lstools/platform_database.py
bridadan: Yes to both! I'll make a little todo in the original post

bridadan:
Great comments! I'll address those in the next round of updates.

One thing I talked with @theotherjimmy offline is making our "subcommands" (mock, list platforms, update) actual command line "subcommands" instead of arguments. I think this PR could probably introduce that concept first and we could address the other commands in a separate PR.

This would give us the ability to fix #251.

bridadan:
After looking into it a bit, argparse doesn't support "optional subparsers". Meaning we can't have `mbedls` list mbeds while still having the `mbedls update` command. We'd need to move to something like [docopt](https://github.com/docopt/docopt) or another CLI library, which should be done in another PR ahead of this. What do you think @theotherjimmy ?

theotherjimmy:
@bridadan Agreed. Add switches for now. We'll prototype mbed-ls 2.0's sub-command style in a future PR.

bridadan:
Ok, I've added quite a bit of output. New platforms are now printed in a table. Updated platform names are printed like `ARM_MPS2_M0 -> ARM_IOTSS_BEID`. An example is shown below:


```
$ mbedls --update
WARNING:mbedls.platform_database:Error loading database C:\Users\bridan01\AppData\Local\mbedls\mbedls\platforms.json: [Errno 22] Invalid argument; Recreating
Fetching platform data from https://os.mbed.com/api/v3/platforms/?format=json...
+-----------+-------------------------------+
| target_id | platform_name                 |
+-----------+-------------------------------+
| 0320      | MTS_MDOT_F411RE               |
| 5001      | ARM_MPS2_M0 -> ARM_IOTSS_BEID |
| 5004      | ARM_CM3DS_MPS2                |
| 7401      | LPC11U35_501                  |
| 7405      | LPC11U35_501                  |
| 9013      | ARCH_LINK                     |
| C000      | LPC812                        |
+-----------+-------------------------------+
The platform database was updated.
```

When no new platforms were found, this is the output:

```
$ mbedls --update
Fetching platform data from https://os.mbed.com/api/v3/platforms/?format=json...
Platform database up-to-date.
```

To accomplish this, I've changed the return value of the `update_from_web` function. The output is detailed in the function's docstring. For errors that occur, it will now raised those as exceptions: `RemotePlatformDataException` for web related issues and `DatabaseUpdateException` for filesystem related issues.

bridadan:
Aaaaand da tests are in 👌

mbed_lstools/platform_database.py
```diff
@@ -480,15 +489,101 @@ def remove(self, id, permanent=False, device_type='daplink', verbose_data=False)
...
+        }
+
+        all_updates = new_platforms.copy()
+        all_updates.update(updated_platforms)
+
+        for id, platform_name in all_updates.items():
             self.add(id, platform_name)
 
-        self._update_db()
+        if not self._update_db():
```
jupe: why update_db does not raise exception ?

mbed_lstools/platform_database.py
```diff
@@ -480,15 +489,101 @@ def remove(self, id, permanent=False, device_type='daplink', verbose_data=False)
...
+
+        If the update fails to contact the provided url, a `ConnectionError` will
+        be raised (part of the `requests` library). If it fails during the web
+        request or JSON parsing, a `RemotePlatformDataException` is raised.
+        If it fails during the write to the filesystem, a `DatabaseUpdateException`
+        is raised.
+
+        Returns dictionary of all updates that occurred in the following format:
+            {
+                "new": {
```
jupe: why this function could not overwrite whole local db? isn’t this try to do same thing anyway..?

mbed_lstools/platform_database.py
```diff
@@ -480,15 +489,101 @@ def remove(self, id, permanent=False, device_type='daplink', verbose_data=False)
...
+            logger.debug(r.text)
+            raise RemotePlatformDataException()
+
+        try:
+            """Expected JSON format is as follows:
+            [
+                {
+                    "productcode": "1234",
+                    "logicalboard": {
+                        "name": "boardname"
```
jupe: is boardname == platform_name ?

mbed_lstools/platform_database.py
```diff
@@ -480,15 +489,101 @@ def remove(self, id, permanent=False, device_type='daplink', verbose_data=False)
...
+                        "old": "<previous platform name>",
+                        "new": "<new platform name>"
+                    }
+                }
+            }
+        """
         r = requests.get(url)
-        platform_data = r.json()
+        try:
+            platform_data = r.json()
```
jupe: would be good if received json blob is schema validated before using json to avoid unexpected behaviour..

jupe:
ups, seems that I was a bit too late for new comments 😁

theotherjimmy:
@jupe It's not released just yet. Comments are still welcome.

mbed_lstools/platform_database.py
theotherjimmy: Next PR?

mbed_lstools/platform_database.py
theotherjimmy: Yes

mbed_lstools/platform_database.py
theotherjimmy: To avoid blowing away the jlink info.

mbed_lstools/platform_database.py
theotherjimmy: I'm not seeing this part of the diff.

bridadan: It never has

mbed_lstools/platform_database.py
bridadan: I do a bit of catching of schema errors on line 541. It prints to debug if the schema changes. I don't think we'll be doing anything stronger than that, except maybe raising a warning?

PR #345, opened 2018-05-11
Add -M, --models-list option to mbedls to listing models

jamesbeyond:
# Description

In order to support FastModels in mbed testing framework, an extra option "-M, --models-list" been added to mbedls.

This option will allow mbedls hooks up with fastmodel_agent module, list all available models in a table

# Details
* "-M" or "--model-list" been added to mbedls
* user could use __mbedls -M \<agent_module>__
  * e.g.  __mbedls -M fastmodel_agent__

Listed models and configuration will be printed out in a table.

This PR is related with other PRs regarding to FastModel support in testing framework 
ARMmbed/greentea#270
ARMmbed/htrun#187

CC. @theotherjimmy @studavekar @bridadan @bulislaw

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -338,6 +338,63 @@ class MbedLsToolsBase(object):
 
         return pt.get_string()
 
+    def list_simulator_module_targets(self, module_name):
+        """! List all models and configs in given simulator module
+        @param module_name is the name of simulator module which be used with greentean/htrun
+        @details module_name e.g. fastmodel_agent
+        @return String with table formatted output
+        """
+        from prettytable import PrettyTable
+
+        try:
+            simulator_module = __import__(module_name)
+        except ImportError as e:
+            logger.error("Could not import module: %s\n%s", module_name, str(e))
+            return ''
+        resource = simulator_module.create()
+        model_dict = resource.list_avaliable_models()
+
+        columns = ['MODEL_NAME', "MODEL_LIB", 'CONFIG_NAME' , 'CONFIG_FILE', 'AVAILABILITY']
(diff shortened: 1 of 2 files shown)


theotherjimmy:
@jamesbeyond I think this is out of scope for mbedls. From the readme:
>detects and lists Mbed Enabled devices connected to the host computer.

This seems like it's something other that a connected device. Also worth noting that the other PRs do not appear to depend on this one.

bridadan:
I'm inclined to agree withe @theotherjimmy for this particular PR. Would you be opposed to allowing the fastmodel_agent to do the listing of the FVPs?

jamesbeyond:
Hi @theotherjimmy @bridadan,
Yes, your right, this RP to mbedls is more separated from the PRs in thrun and greentea.

The reason I put it here becasue:
1. in current design, fastmodel_agent module dose not having any CLI interface. It only can be accessed used via greentea/mbedls.

2. The purpose for mbed-ls tools is to listing the targets which are able to run mbed os. In this context, I thought models also belong to there, they just special category of targets compare the physical hardware.

So If you really think this function not fits mbed-ls' purpose, I can move these to fastmodel_agent. and enable a CLI there.

theotherjimmy:
> I can move these to fastmodel_agent. and enable a CLI there.

That's my personal preference.

If you want to change mbed-ls's scope to include "anything that can run mbed-os", we could include this. I would strongly prefer that it list fast models with the default command, no switch.

jamesbeyond:
OK, then I'll move the function of listing available models to the fastmodel_agent itself @theotherjimmy

theotherjimmy:
Thanks. This PR is no longer needed.

PR #347, opened 2018-05-30
Release Version 1.4.5

theotherjimmy:
# Fixes
 * Correct corner case that resulted in unmounted targets being returned
 when they were not requested


theotherjimmy:
Gotta merge stuff first.

PR #349, opened 2018-05-31
Revert "Adding a --update command for pulling new target defintions from online"

theotherjimmy:
Reverts ARMmbed/mbed-ls#341


theotherjimmy:
@bridadan Revert this PR, then add changes when you want to add --update.

PR #350, opened 2018-05-31
Release Version 1.4.5

theotherjimmy:
# Fixes
  * #335 - Patch edge case for devices with unreadable files
  * #346 - Silence warning about htm-based target id
  * #339 - Warn if device missing from Platform DB

 # New Targets
  * #340 - Platform DB: Add NRF51_DK to jlink section.


theotherjimmy:
@jupe links and all

PR #354, opened 2018-06-05
Adding directory entries to platform data

### LINKED ISSUE #336: Indicating about ASSERT.TXT
In CI we need to know if interface asserts and ASSERT.TXT appears to mount point. Would be optimal solution if mbed-ls could indicate if ASSERT.TXT file exists when list boards. mbed-ls could e.g. read that file when it exists and include it in board json, or just report all files in mount point so that application could read them when needed. This would help to avoid using bad state boards in CI.

bridadan:
This fixes #336


jupe:
is this coming for next patch rel ?

bridadan:
I see no reason why it can't!

PR #369, opened 2018-07-16
Vid pid details

bridadan:
This is part of the solution to #165. The next step would take place in the base class which would be responsible for altering the `device_type` field.

This PR implements a mechanism for each OS (Windows, Linux, and Mac) to provide the Vendor ID and Product ID of the Mbed devices present on the system. These two data items are present in the results as `vendor_id` and `product_id` respectively.

**Both the `vendor_id` and `product_id` fields should be displayed as a 4 character, lower case hex string.**

FYI @theotherjimmy @jupe 

Here is some sample output:

```
$ mbedls -j
[
    {
        "daplink_build": "Jan 11 2016 16:12:36",
        "daplink_url": "http://mbed.org/device/?code=07400221076061193824F764",
        "daplink_version": "0221",
        "device_type": "daplink",
        "directory_entries": [
            "DETAILS.TXT",
            "MBED.HTM"
        ],
        "mount_point": "/mnt/pci-0000_00_06_0-usb-0_3_1_1-scsi-0_0_0_0",
        "platform_name": "NUCLEO_F411RE",
        "platform_name_unique": "NUCLEO_F411RE[0]",
        "product_id": "374b",
        "serial_port": "/dev/ttyACM1",
        "target_id": "07400221076061193824F764",
        "target_id_mbed_htm": "07400221076061193824F764",
        "target_id_usb_id": "0671FF554856805087112815",
        "vendor_id": "0483"
    },
    {
        "daplink_auto_reset": "0",
        "daplink_automation_allowed": "1",
        "daplink_bootloader_crc": "0xb92403e6",
        "daplink_bootloader_version": "0244",
        "daplink_daplink_mode": "Interface",
        "daplink_git_sha": "0beabef8aa4b382809d79e98321ecf6a28936812",
        "daplink_hic_id": "97969900",
        "daplink_interface_crc": "0x434eddd1",
        "daplink_interface_version": "0246",
        "daplink_local_mods": "0",
        "daplink_overflow_detection": "1",
        "daplink_remount_count": "0",
        "daplink_unique_id": "0240000032044e4500257009997b00386781000097969900",
        "daplink_usb_interfaces": "MSD, CDC, HID",
        "daplink_version": "0246",
        "device_type": "daplink",
        "directory_entries": [
            "MBED.HTM",
            "DETAILS.TXT"
        ],
        "mount_point": "/mnt/pci-0000_00_06_0-usb-0_2_1_0-scsi-0_0_0_0",
        "platform_name": "K64F",
[...]

Issue #165, opened 2017-03-08
identify ST boards firmware version correct way


jupe:
Seems that ST boards or any dev kit using ST-Link firmware need to be identified differently than DAPLink boards (see STLink utility tool). Now mbed-ls gives you incorrect "daplink_version" number and makes confusion.

bridadan:
@jupe True, it does report "daplink_version" for all boards, even if they aren't based on DAPLink. May I ask how you're using this information?

jupe:
in test beds we want to know what firmware version we have in boards, and update if there is old ones...

jeromecoutant:
Hi

Maybe I can propose STM32 ST-LINK utility software from
http://www.st.com/en/development-tools/stsw-link004.html

ST-LINK_CLI.exe -List
will give you the list of all connected STM32 boards

ex:
--- Available ST-LINK Probes List ---
ST-LINK Probe 0:
     SN: 066BFF545253675187212436
     FW: V2J27M15

ST-LINK Probe 1:
     SN: 0672FF504852897267175941
     FW: V2J29M18

I tried few commands to associate SN with a board.
I found this one.

If you have only 1 board connected:

$ ./ST-LINK_CLI.exe -c
ST-LINK SN : 066BFF545253675187212436
ST-LINK Firmware version : V2J27M15
Connected via SWD.
SWD Frequency = 4000K.
Target voltage = 3.3 V.
Connection mode : Normal.
Device ID:0x458
Device family :STM32F410xx

If your different boards are from different family, you can keep them connected and indicate SN like:

$ ./ST-LINK_CLI.exe -c SN=066BFF545253675187212436
ST-LINK SN : 066BFF545253675187212436
ST-LINK Firmware version : V2J27M15
Connected via SWD.
SWD Frequency = 4000K.
Target voltage = 3.3 V.
Connection mode : Normal.
Device ID:0x458
Device family :STM32F410xx

theotherjimmy:
@jeromecoutant And what do you have for a Linux host?

ciarmcom:
ARM Internal Ref: MBOTRIAGE-481

bridadan:
This can be closed now!

PR #370, opened 2018-07-17
VID/PID based device type detection

bridadan:
This builds on #369 to solve issue #165.

In PR #299 I added device type detection. It used a strategy that made use of the files present on the device to identify which firmware was most likely running on the interface chip. This was good enough to differentiate between JLink and DAPLink based firmwares, but got a little more complicated for firmwares like ST-Link. ST-Link tends to mimic DAPLink files quite closely, so this option isn't really feasible any more.

This PR switches the strategy to instead identify device types by the Vendor ID. In this case I'm not currently using the Product ID as it does not seem necessary at this point. There also is a small chance that the Product IDs may vary a bit depending on the release of the interface firmware, so I figured I'd postpone this change until we have evidence to the contrary.

## Details to review (please 😄)
This also adds `stlink` to the list detectable firmwares. The STLink detection makes use of the `DAPLink` update logic at the moment, meaning you will see `daplink_build` and `daplink_version` in the JSON output, even though this is using stlink. I did this on purpose because this is how the information would have showed up in earlier versions. Should I also provide this info as `stlink_*`? Should I leave it how it is?

FYI @jupe @theotherjimmy

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -223,7 +223,7 @@ class MbedLsToolsBase(object):
             {
                 'daplink': self._update_device_details_daplink,
                 'jlink': self._update_device_details_jlink
-            }[device['device_type'] or 'daplink'](device, read_details_txt, directory_entries)
+            }[device['device_type'] or 'daplink'](device, read_details_txt)
         except (OSError, IOError) as e:
             logger.warning(
                 'Marking device with mount point "%s" as unmounted due to the '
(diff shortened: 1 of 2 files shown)


mbed_lstools/lstools_base.py
```diff
@@ -69,6 +69,12 @@ class MbedLsToolsBase(object):
     DETAILS_TXT_NAME = 'DETAILS.TXT'
     MBED_HTM_NAME = 'mbed.htm'
 
+    VENDOR_ID_DEVICE_TYPE_MAP = {
+        '0483': 'stlink',
```
jupe: Can we rely on to these thtee VID in every boards now? what happens if VID does not match to some?

bridadan: Yes I believe we can, especially on anything released within the past few years. If the VID does not match it will default to the 'daplink' style detection (just like it used to).

PR #375, opened 2018-07-26
Platform name api

### LINKED ISSUE #365: API to get supported platforms names
Current only way to fetch all supported target names are like this:

```
mbeds = mbed_lstools.create()
db_items = list(mbeds.plat_db.items())
supported_targets = sorted([i[1]["platform_name"] for i in db_items])
```

Like you see it's not very simple and clear way. I would like to have dedicated public API to achieve same, eg `mbeds.get_supported_targets()` which could do these tricks for me.

bridadan:
This one fixes #365.

FYI @jupe

### DIFF
--- a/mbed_lstools/lstools_base.py
+++ b/mbed_lstools/lstools_base.py
@@ -403,6 +403,18 @@ class MbedLsToolsBase(object):
         self.dummy_counter[platform_name] += 1
         return platform
 
+    def get_supported_platforms(self, device_type=None):
+        """! Return a dictionary of supported target ids and the corresponding platform name
+        @param device_type Filter which device entries are returned from the platform database
+        @return Dictionary of { 'target_id': 'platform_name', ... }
+        """
+        kwargs = {}
+        if device_type is not None:
+            kwargs['device_type'] = device_type
+
+        items = self.plat_db.items(**kwargs)
+        return {i[0]: i[1] for i in items}
+
     @deprecated("List formatting methods are deprecated to simplify the API. "
                 "Please use 'list_mbeds' instead.")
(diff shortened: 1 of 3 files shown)


mbed_lstools/lstools_base.py
```diff
@@ -403,6 +403,18 @@ def get_dummy_platform(self, platform_name):
         self.dummy_counter[platform_name] += 1
         return platform
 
+    def get_supported_platforms(self, device_type=None):
+        """! Return a dictionary of supported target ids and the corresponding platform name
+        @param device_type Filter which device entries are returned from the platform database
+        @return Dictionary of { 'target_id': 'platform_name', ... }
```
jupe: is there some needs for target_id’s? Isn’t it enough for our tools to get just platform_names? (at least mbed-flasher does not need target-ids) function names says ”platforms” so at least I would expect to get just platform_names in array 🤔

mbed_lstools/lstools_base.py
```diff
@@ -403,6 +403,18 @@ def get_dummy_platform(self, platform_name):
         self.dummy_counter[platform_name] += 1
         return platform
 
+    def get_supported_platforms(self, device_type=None):
```
jupe: thanks for device type filtering - mbed-flasher will need it 👍

mbed_lstools/lstools_base.py
bridadan: I mostly did it to not create two APIs for fetching platform names and target ids.

jupe: That’s fine for me.

PR #390, opened 2018-09-05
Add support for XPLAINED PRO SAML21 board

janjongboom:
Prerequisite for landing ArmMbed/mbed-os#7276. Adding support for Atmel XPLAINED Pro development board. Atmel now uses something similar as DAPLink for programming. Note that the interface chip might need to be updated to latest version to support drag-n-drop programming (and thus mbed-ls), you're prompted to do this when opening Atmel Studio and connecting to the board the first time.

@theotherjimmy

### DIFF
--- a/mbed_lstools/__init__.py
+++ b/mbed_lstools/__init__.py
@@ -36,3 +36,4 @@ line tool (see below).
 
 from .main import mbedls_main
 from .main import create
+
--- a/mbed_lstools/darwin.py
+++ b/mbed_lstools/darwin.py
@@ -28,7 +28,7 @@ logger.addHandler(logging.NullHandler())
 DEBUG = logging.DEBUG
 del logging
 
-mbed_volume_name_match = re.compile(r'\b(mbed|SEGGER MSD)\b', re.I)
+mbed_volume_name_match = re.compile(r'\b(mbed|SEGGER MSD|ATMEL EDBG Media)\b', re.I)
 
 def _find_TTY(obj):
     ''' Find the first tty (AKA IODialinDevice) that we can find in the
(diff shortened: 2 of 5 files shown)


theotherjimmy:
@janjongboom You have a merge conflict here. Let me see if it's easy to resolve.

theotherjimmy:
@janjongboom It looks like the merge conflict is non-trivia. Do you have time to resolve it?

martinichka:
@janjongboom great work! Will it work and it is tested on Linux as well? I can support with some testing/debugging on Linux if needed.

janjongboom:
@martinichka It should but haven't verified it. Will (hopefully) do it when back in Netherlands later this week.

janjongboom:
@bridadan @theotherjimmy 

Tested on macOS 10.13, Ubuntu 16.04 and Windows 7, all works. Have rebased against master.

janjongboom:
@bridadan Not sure how to add test data for this board, as I'm not using the content but rather the USB ID from the device.

theotherjimmy:
@janjongboom I'm taking this in good faith for now. I would like to get one of these boards and scrape that testing data.

bridadan:
Thanks again for doing all of this work! You did all of the hard parts 😄 

> @bridadan Not sure how to add test data for this board, as I'm not using the content but rather the USB ID from the device.

That's a good point! Though as a matter of good practice and to help us in the future, I'm hoping to collect the files present on at least one platform of each type of debugger interface. This is so we don't have to go track down the board each time we want to test to make sure we didn't break anything. Instead we can just simulate the device in its entirety. This is really helpful when we're trying to improve the detection algorithm.

So if you could just copy and paste the contents of the "drive" that shows up when you plug in the board into the `test_data` directory (https://github.com/ARMmbed/mbed-ls/tree/master/test/test_data), that would help us out a lot in the future!

janjongboom:
@bridadan OK, will do (but in 2 weeks because I didn't bring the board :-)).

PR #404, opened 2018-11-22
Add ARM_MUSCA_A1 target

gaborkertesz:
This commit adds ARM_MUSCA_A1 Cortex-M33 based target with 5006 ID.

mbed-os pull request adding this new platform:
https://github.com/ARMmbed/mbed-os/pull/8845


ashok-rao:
@bridadan , @theotherjimmy , @deepikabhavnani : Could you please review? Thanks.

ashok-rao:
@bridadan : This should be fixed now. Could you please re-check? Also, there is an identical A1 and A2 ..so just to differentiate, we kept the slug and name identical but changed the target name for the compiler.. hope this helps?

PR #417, opened 2019-01-08
Update platform_database.py

RomanSaveljev:
Is this too bad?


bridadan:
@RomanSaveljev and I were working on this together offline, we will revisit this.

